PR #6, opened 2022-05-19
[WIP] testing if we can run docker registry as service during action

vsoch:
Before adding tests, I need to figure out a good testing strategy for testing the client interactions with a registry, so my first idea is to try having a docker registry run as a service in the action. This PR is a WIP to the main development branch and y'all can largely ignore it.


vsoch:
yay! Okay this should work for testing! I'll do that next, probably later tomorrow.

image

Issue #5, opened 2022-05-18
Dinosaur Todos


vsoch:
The development branch (with un-authenticated push and pull working!) is at https://github.com/oras-project/oras-py/tree/design-two, and this is my list of TODO before it can be considered for review (e.g., the first PR to main).

 - [x] add same views with auth
 - [x] docstrings
 - [x] release workflow https://github.com/oras-project/oras-py/pull/9
 - [x] pretty (branded) documentation https://github.com/oras-project/oras-py/pull/9
 - [x] logout command
 - [x] add testing https://github.com/oras-project/oras-py/pull/6 
 - [x] isort, mypy, pyflakes black https://github.com/oras-project/oras-py/pull/6
 - [x] ensure we can push/pull directory, should be tar-d and extracted https://github.com/oras-project/oras-py/pull/8
 - [x] add typing
 - [x] add schemas for manifest, annotations, etc. https://github.com/oras-project/oras-py/pull/8
 - [x] we should have common function to parse errors in json 'errors' -> list -> message https://github.com/oras-project/oras-py/pull/8
 - [x] container bases and automated builds https://github.com/oras-project/oras-py/pull/7

And these need further discussion about design, etc, but I don't think are necessarily blockers for getting a community version for people to start contributing to!

 - [x] add example (custom) GitHub client (talking with a colleague about this - it will be the first "here is how to use the API" example and I'll write docs for that then.
 - [ ] difference between plain_http and insecure (I just used the latter)
 - [ ] we haven't added path traversal, or cacheRoot to pull
 - [ ] environment variables like `ORAS_CACHE` - what others?
 - [ ] refactor internals to be more like oras-go (e.g., provider, copy?) I think I want to be convinced this abstraction is useful.
 - [ ] should there be a tags function? 
 - [ ] need to have git commit, state, added to defaults on install/release. See [here](https://github.com/oras-project/oras/blob/main/Makefile). - I'd rather not have this, I don't like when version/install relies on pinging git.

For some background, I made the unwise decision the first time around to mimic the Go client, and that was a bad idea :) This time around I'm implementing everything straight forward (e.g, push and pull) and then I'll add abstraction to make it easy to extend. I am curious about what cases of "copy" have been useful for users or developers? Conceptually it's much more clear with push and pull (and I suspect what 80% of users want!)

vsoch:
[...]

PR #10, opened 2022-05-20
First Draft Implementation of Oras Python!

vsoch:
Hey everyone! I've finished up the "what I think are hard requirements" from #5 and here is a summary of the first round of work that will provide a working oras client. Some notes:

 - Basic support for login/logout/push/pull with and without authentication
 - Artifacts can be single blobs or directories, the latter of which get a different media type and are tar-guzzed up and extracted on download. We get the filename from an annotation.
 - Basic tests for linting and core functionality is added, with the exception of authenticated views. I don't have an easy way to set up a full auth server in an action, so the most we could do right now is to add a workflow that brings up a server with a password and then we test login/logout. I'm happy to do that in another PR (after this one) and if it's desired give me a +1 and I'll open an issue.
 - Functions have docstrings and parameters defined, along with typing
 - We lint-all-the-things, including black, pyflakes, mypy, and isort. 
 - A Makefile makes install / lint easy for the developer
- There is a release workflow added already with a credential to pypi, so making a release of the software here releases to pypi.
- I've made pretty branded docs that match oras.land (and they link there) and I'd like to extend this to include Python docstrings, just want to play around with some ideas before just defaulting to standard sphinx (will open issue after merge here). Oh! And would be super awesome to get the docs previews hooked up to Netlify? I love Netlify but none of my academic projects use it, because the free tier is too small and we cheap. :laughing: 
- Testing is for functionalty and linting, of course
- Manifests and other json dumps are validated with jsonschema
- An automated container build to deploy to ghcr.io/oras-project/oras-py:latest (or a tag on release)

And there are still a list of things to discuss under #5 - e.g., some details that aren't yet added from oras-go (e.g., envars to drive things) that I think would be best if someone familiar could go through and add a list to an issue, and I'll tackle more slowly over time. I do think it's important to get something up sooner than later so the community can give feedback and contribute!

With respect to "copy" - I know the original oras didn't have this design and it was refactored, but I'm not convinced yet that this abstraction is useful. That said, if someone comes to me and says "I need this to be a more abstract storage that I can copy to" we can refactor to generalize. My spidey sense is that 90% of people just want to push/pull from standard registries, and so I followed the logic of "keep it simple, stupid" and implemented that.

After this original round of work I'm going to be working with @wolfv to discuss his needs to use oras with Mamba/Conda, and make sure the library here meets them. I suspect we will be putting together the first "custom client example" once we have that chat (and we can open a separate issue for this too).

Yay! :partying_face: Sorry I should have done this from the getgo, I wasted a lot of time trying a really bad idea, sort of assumed someone else would step up, and realized two days ago nobody was so I'd just hammer it out. :hammer:


sajayantony:
Now we need python reviewers :) But I'm in favor or merging into main first and incrementally addressing things through issues or smaller PRs.

vsoch:
I'm good with that too @sajayantony ! I am here at your service for all the issues / feature requests that come in, and I'm going to keep working on the things I noted above, and probably think of more as I go. I'm not experienced with oras-go but it would be fairly easy to look there and get a sense of what I'm missing, environment variable and interaction wise. I also want to get that GitHub example up with developer docs (for using the Python API) sooner than later.

Happy Thursday!

sajayantony:
@jdolitsky @deitch

oras/client.py
```diff
@@ -0,0 +1,139 @@
...
+    This is intended for controlled interactions. The user of oras-py can use
+    this client, the terminal command line wrappers, or the functions in main
+    in isolation as an internal Python API. The user can provide a custom
+    registry as a parameter, if desired. If not provided we default to standard
+    oras.
+    """
+
+    def __init__(
+        self,
+        hostname: str = None,
```
pbarker, line 28: Optionals?

oras/logger.py
```diff
@@ -0,0 +1,313 @@
...
+
+    @property
+    def is_tty(self) -> bool:
+        """
+        Determine if we have a tty environment
+        """
+        isatty = getattr(self.stream, "isatty", None)
+        return isatty and isatty()  # type: ignore
+
+    def emit(self, record):
```
pbarker, line 64: type?

oras/logger.py
```diff
@@ -0,0 +1,313 @@
...
+        Handle a log message.
+
+        Arguments
+        ---------
+        msg : the message to handle
+        """
+        for handler in self.log_handler:
+            handler(msg)
+
+    def set_stream_handler(self, stream_handler):
```
pbarker, line 136: type?

oras/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+annotations = {
+    "type": "object",
+    "patternProperties": {
+        "\\w[\\w-]*": {"type": "string"},
+    },
+}
+
+# Layer
+
+layerProperties = {
```
pbarker, line 19: may be fun to generate these in the future

oras/tests/artifact.txt
```diff
@@ -0,0 +1 @@
+hello dinosaur
```
pbarker, line 1: 🦕

oras/utils/fileio.py
```diff
@@ -0,0 +1,262 @@
...
+    """
+    Make a targz (compressed) archive from a source directory.
+    """
+    dest_name = dest_name or get_tmpfile(suffix=".tar.gz")
+    with tarfile.open(dest_name, "w:gz") as tar:
+        tar.add(source_dir, arcname=os.path.basename(source_dir))
+    return dest_name
+
+
+def extract_targz(targz, outdir):
```
pbarker, line 31: type

pbarker (commented):
Beautiful code! Thanks for doing this!

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:22.04
+
+# docker build -t oras-py .
```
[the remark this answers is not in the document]
vsoch: Yes I always comment my Dockerfile with the build command for the developer user to quickly see the intended build.

oras/auth.py
```diff
@@ -0,0 +1,87 @@
...
+import base64
+import os
+import re
+from typing import List, Optional
+
+import oras.utils
+from oras.logger import logger
+
+
+def load_configs(configs: List[str] = None):
```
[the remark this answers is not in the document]
vsoch: In theory yes, although we always provide from the client so it always comes as a list. But I'm happy to tweak this to be Optional so it's more correct!

vsoch:
@pbarker thank you for the review! I've fixed everything locally and will likely do a push with changes after I worked a bit more on the requests in #11 (which has some great ideas that I think will make it better!)

sajayantony (approved):
LTGM to merge and iterate on this going forward.

vsoch:
Thank you @sajayantony ! Let's hold off merging until I get the final tweaks in this evening, if that's ok :)

sajayantony:
@vsoch - If there is one thing I have learned, it is "don't merge too fast"  :) I've been burnt a few times.

vsoch:
I'm going to wait for this to be merged until I start playing around with a documentation strategy - this PR is getting too chonky (well it's already superchonk) and I'm mostly worried about messing something up.

wolfv:
This looks great! I would be happy to help maintain the `oras-py` client going forward. Just to give some context, we're looking into using the Github OCI registry as mirror / backup and we want to offer "native" support in the mamba package manager asap.

vsoch:
yay! That sounds super fun! :partying_face: I'm happy to help maintain too, and continue tweaking for your use case (I'm a huge fan of mamba/conda btw!) 

And I just added spell checking, contributor stats, and some super pretty docs - they match the oras.land branding but of course have more detailed usage (and docstrings) for the Python API here. I also updated blob -> blob_path, not the perfect solution to your issue but works for the time being.

image

And here is what the rendered docstrings look like. The "sphinx format" isn't as pretty to see in the text imho (and it's redundant to have the type again) but it's a solution for now until/unless we decide on something better. I do think a Python API should have the basic usage / source readily available so people don't have to search through GitHub.

image

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+LABEL MAINTAINER @vsoch
+ENV PATH /opt/conda/bin:${PATH}
+ENV LANG C.UTF-8
+RUN apt-get update && \
+    apt-get install -y wget && \
```
wolfv, line 9: coming from the mamba world I'd advocate for the `micromamba` docker image here, and using the `python` packages from the `conda-forge` channel: https://github.com/mamba-org/micromamba-docker/ – you could also install `ipython` from conda-forge :)

vsoch: Haha, fair! We could probably add another recipe for that if you think it’s much better, even elsewhere. My preference generally is for more vanilla base images but you are right I broke that rule a bit with conda instead of installing python3. We can have a discussion with other maintainers to gauge preferences - I’d be ok with any of these ideas really!

wolfv: I guess if you want to use `pip` anyways, we could also use the official `python` base images? (https://hub.docker.com/_/python)

wolfv:
I think this looks great. Should it be merged?

I would be happy to try it out in our OCI mirroring actions (we're currently calling `oras` in a python subprocess...). This will be much more convenient.

sajayantony (approved):
Initial draft for ORAS-py

vsoch:
Woot! If I'm allowed to merge my own PRs (at least for this draft) let's do it. I'll follow up quickly if there is any issue with the docs build, etc.

Issue #11, opened 2022-05-20
Features that we like in conda_oci_mirror


wolfv:
Hi @vsoch,

awesome work on the initial implementation. I think I have some feedback based on our experience implementing something like this for `conda-oci-mirror`: https://github.com/mamba-org/conda_oci_mirror

We currently push three layers per "conda"-package to the OCI registry. The first layer is the `.tar.bz2` package file. The second layer is the `index.json` file, containing the metadata and dependencies of the package (useful for creating an index over the whole repo). The third file is a `.tar.gz` of the `info/...` directory that every conda package contains (that directory contains a list of all files of the package and more interesting metadata that would be nice to access without downloading the whole package).

In our client, we added some easy-access functions for the latter two objects. For the `index.json` we first fetch the manifest and look for the layer with the appropriate mediaType (`vnd.conda.index+json` or something like that) and then pull only that layer blob and immediately parse it to json using `res.json()` from `requests`.

For the `.tar.gz` file we pass it on to the `tarfile` reader to have access to all files inside in a nice way.

Maybe exposing the `get_blob` without an underscore could be cool?

I thought having a `Layer` class (with a `.to_json()` function) might be nicer than the `NewLayer` free function that returns a dict (this is just a nitpick, I guess)

We also have [a function to retrieve all tags](https://github.com/mamba-org/conda_oci_mirror/blob/f300085bdf7725b2d134f2c48db415d7129691fa/conda_oci_mirror/oci.py#L49-L75
) for a given container. This is necessary for efficient mirroring and could also fit well into `oras-py` I think.
 
Anyways, this looks super promising, thanks for your work!!

wolfv:
Btw. just in case you wanted to see one of our 'packages' as described here on the OCI registry: https://github.com/orgs/channel-mirrors/packages/container/package/conda-forge%2Fosx-64%2Fxtensor

vsoch:
hey @wolfv ! It looks like the request to get a manifest returns a 500 error? I tried interacting with:

Regular oras:

```
$ oras pull -a ghcr.io/channel-mirrors/conda-forge/osx-64/xtensor:0.9.0-0
Error: unexpected status code [manifests 0.9.0-0]: 500 Internal Server Error
```

docker manifest inspect

```
$ docker manifest inspect ghcr.io/channel-mirrors/conda-forge/osx-64/xtensor:0.9.0-0
received unexpected HTTP status: 500 Internal Server Error
```
[...]

PR #12, opened 2022-05-20
Review fixes and example

vsoch:
This PR is going to address PR review in #10, along with:

 - adding an example for a custom client (conda mirror) that has three artifacts in a manifest, and does custom parsing based on the content type.
 - adds a decorator ensure_container that allows for calling any registry function with a string container name OR already parsed container. 
 - add missing typing for Optional, etc.
 - adds an endpoint to retrieve tags metadata
 - Implements Layer as a class, still with NewLayer to instantiate and return the Layer.to_dict() if we want to skip having the class.
 - get_blob is now a public function
 - fixing a bug in the manifest schema validation

which were discussed in #11. Likely we will continue iterating on 11 to make it into more of a desired class with functions over a demo to show how things work!

### DIFF
--- a/oras/oci.py
+++ b/oras/oci.py
@@ -4,7 +4,7 @@ __license__ = "Apache-2.0"
 
 import copy
 import os
-from typing import Dict, Tuple
+from typing import Dict, Optional, Tuple
 
 import jsonschema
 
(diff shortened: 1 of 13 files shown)


oras/oci.py
```diff
@@ -44,22 +44,43 @@ def get_annotations(self, section: str) -> dict:
         return {}
 
 
-def NewLayer(blob, media_type: str = None, is_dir: bool = False) -> dict:
+class Layer:
+    def __init__(
+        self, blob: str, media_type: Optional[str] = None, is_dir: bool = False
```
wolfv, line 49: just to confirm -- blob is a path here? I think it could be confusing to users wether blob is a path to a file or a string to upload.
If we'd use `pathlib.Path` as type, it would probably be pretty obvious.

However, we could also use a `io` type as input (so that the user can give a File, StringIO, or BytesIO object).

Just some additional thoughts. 

We could also have named functions `Layer.from_string(...)`, `Layer.from_blob(BytesIO(...))` etc.

PR #19, opened 2022-05-25
updating oras project site metadata and disabling contributors workflow

vsoch:
the contributors workflow currently is done by a github bot and is not a signed commit so I am disabling it until we discuss!
I'm also update the docs site metadata to be for here - I used a template I made for River and just totally forgot it was there. :laughing: 

### DIFF
--- a/.github/workflows/update-contributors.yaml
+++ b/.github/workflows/update-contributors.yaml
@@ -3,7 +3,7 @@ name: Update Contributors
 on:
   push:
     branches:
-      - main
+      - doesnotexist
 
 jobs:
   Update:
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):

.github/workflows/update-contributors.yaml
```diff
@@ -3,7 +3,7 @@ name: Update Contributors
 on:
   push:
     branches:
-      - main
+      - doesnotexist
```
sajayantony, line 6: What is the purpose of the branch change?

vsoch: Right now the all contributors workflow checks for new contributors and updates the metadata file. However because the commit isn't signed, it will never pass CI (see https://github.com/oras-project/oras-py/pull/18) so I'm disabling it to run for now (via a branch that doesn't exist) so we can discuss how to either fix it, allow an exception, or remove it.

sajayantony (approved):

PR #21, opened 2022-05-25
adding auth tests

vsoch:
this will add a workflow to test basic auth, meaning we can login and logout. There is also a current bug with using login/logout because it was not tested! I am also adding brief docs for how to use these functions within Python


sajayantony (approved):

vsoch:
Thank you! Let's get these fixes in asap.

PR #27, opened 2022-05-26
adding 1_python.md site for oras-www

vsoch:
What we can do here is add a page here, and then setup a workflow in oras-www to receive a dispatch event to do a build when it is triggered. I can first test this locally (and it will open a PR) and when we like that, we can add the dispatch event to be triggered in oras-py, and then push directly given that the commit is signed. Given that netlify is set up to build on push to main, if this
works it will mean that we always trigger an update and rebuild on oras-www when there is an update to oras-py.

So plan of action, more explicitly:

1. review and merge this page here
2. add a dispatch event to oras-www that will retrieve the updated file, put it where it needs to be, and pull request to main
3. test this with a local ping to see that it works
4. add the dispatch event to trigger from a merge to main for oras-py
5. when all that is working, change from PR to push
6. then update the oras-py docs to not include the redundant information and link to oras.land
 
### DIFF
--- a/docs/1_python.md
+++ b/docs/1_python.md
@@ -110,8 +110,8 @@ Note that oras-py will not remove content from your docker config files, so
 there is no concept of a "logout" unless you are using the client interactively,
 and have configs loaded, then you can do:
 
-```bash
-$ cli.logout(hostname)
+```python
+cli.logout(hostname)
 ```
 
 ### Push
--- a/docs/1_python.md
+++ b/docs/1_python.md
@@ -1,0 +1,197 @@
+# Oras Python User Guide
+
+Oras Python will allow you to run traditional push and pull commands for artifacts,
+or generate a custom client. This small user guide will walk you through these various
+steps, and please open an issue if functionality is missing.
+
+## Installation 
+
+### Pypi
+
+The module is available in pypi as [oras](https://pypi.org/project/oras/)
+and you can install as follows:
+
+```bash
+$ pip install oras
+``` 
+   
+You can also clone the repository and install locally:
+
+```bash
+$ git clone https://github.com/oras-project/oras-py
+$ cd oras-py
+$ pip install .
+``` 
+
+Or in development mode:
+
+```bash
+$ pip install -e .
+```
+
+Development mode means that the install is done from where you've cloned the library,
+so any changes you make are immediately "live" for testing.
+
+### Docker Container
+
+We provide a [Dockerfile](https://github.com/oras-project/oras-py/blob/main/Dockerfile) to build a container with the client.
+
+```bash
+$ docker build -t oras-py .
+
+$ docker run -it oras-py                                                                                                                   
+# which oras-py
+/opt/conda/bin/oras-py
+```
+
+## Registry
+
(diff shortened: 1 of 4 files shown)


vsoch:
And removed pytest-runner dependency per discussion in https://github.com/conda-forge/staged-recipes/pull/19075#discussion_r882995199

vsoch:
Sorry to keep bothering you for reviews @sajayantony - I want to get everything right for this first draft, including the docs where they need to be! I really appreciate it.

SteveLasker:
adding @shizhMSFT for 👀

docs/1_python.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Login
+
+Once you create (or already have) a registry, you will want to login. You can do:
+
+```bash
+$ oras-py login -u myuser -p mypass localhost:5000
+
+# or localhost (insecure)
+$ oras-py login -u myuser -p mypass -k localhost:5000 --insecure
```
shizhMSFT, line 86: Note that `--insecure` means skipping certificate verification for TLS while `--plain-http` means `HTTP` instead of `HTTPS` for connections.

Are we using the correct flags?

shizhMSFT: BTW, what is the full option name for `-k`? or is it a typo? Note that `-k` is not there at line 83.

docs/1_python.md
```diff
@@ -0,0 +1,197 @@
...
+$ echo mypass | oras-py login -u myuser -k localhost:5000 --insecure --password-stdin
+Login Succeeded
+```
+
+Note that oras-py will not remove content from your docker config files, so
+there is no concept of a "logout" unless you are using the client interactively,
+and have configs loaded, then you can do:
+
+```bash
+$ cli.logout(hostname)
```
shizhMSFT, line 114: It does not look like a bash command. Should it be a block of python script?

docs/1_python.md
vsoch: @shizhMSFT it's likely not, and this is noted in #5. This would be a separate issue from the documentation here, that when fixed we'd update it here too. The current scope of work for this PR is to get the current docs deployed on oras.land.

vsoch: `-k` is insecure https://github.com/oras-project/oras-py/blob/0521bd85a1b544cf3bb3132aee9e7f1d3b16f29c/oras/cli/__init__.py#L133-L140

but should be `--keep-old-files` with https://github.com/oras-project/oras/blob/3149134ede7e0e065a34e762b9ff444d82803305/cmd/oras/pull.go#L73 (it looks like a typo).

vsoch: I can fix that here, give me a minute.

docs/1_python.md
vsoch: Yes good catch!

docs/1_python.md
vsoch: okay -k is fixed, new issue opened for clarifying the flags https://github.com/oras-project/oras-py/issues/29

shizhMSFT (approved):
LGTM

vsoch:
Thank you!

PR #28, opened 2022-05-27
Docker optional dependency

vsoch:
We currently are just using it for login and finding configs, and since osx on conda does not have it (and we want to support it) it makes sense to have docker be optional, meaning we fall back to manual ways to do the same thing. I am leaving it as an extra variant for now in case someone prefers using docker, however if we have confidence it is not needed it can be removed entirely.

See https://github.com/conda-forge/oras-py-feedstock/issues/1 and ping @wolfv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - docker is an optional dependency, to minimize dependencies (0.0.12)
  - Adding authenticated login tests, fixing bugs with login/logout (0.0.11)
  - First draft release with basic functionality (0.0.1)
  - Initial skeleton of project (0.0.0)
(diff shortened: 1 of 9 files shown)


wolfv:
Awesome! I think the regular OS X 64 on conda-forge has it, just the osx-arm64 (M1) chip not yet.

vsoch:
Ah gotcha! Also I was testing newer versions of docker (on Linux/Ubuntu) and the function behavior wasn't consistent (e.g., updating broke some functions) so that is what first incentivized me to want to remove it / make it optional. Likely we will need a user that has a more complicated docker config setup to know if/what is missing.

sabre1041 (approved):
LGTM

vsoch:
Thank you! I think we will need an admin approval (ping @sajayantony!) before the merge is allowed (and I'm not too keen on clicking the "merge without waiting for requirements to be met" box.

When this goes in, I'm planning to release after #32 is merged since it fixes some bugs with respect to pushing to GitHub packages. Then we will have a new release for conda-forge too!

vsoch:
@wolfv I've just triggered a release, so I'll follow up on conda-forge to update the recipe to not require docker.

PR #41, opened 2022-07-07
refactoring docs to markdown instead of rst

### LINKED ISSUE #40: Convert rst to markdown :)

vsoch:
The drawback is that we lose the nice toctree offered by rst, and instead need to hard code the sidebar in the conf.py
as a variable. This should be OK given that we do not want to hugely change the structure of the site. We are essentially
choosing user ease of editing markdown over automatically generated sidebars.

This will fix #40 


vsoch:
Ah! I think I fixed the bug with the Toctree! Going to change this to a WIP until I update it.

docs/getting_started/developer-guide.md
```diff
@@ -0,0 +1,248 @@
...
+    # Read info.json media type, find correct blob...
+    # download correct blob, etc.
+```
+
+Specifically the function `self.get_blob` will allow you to do that.
+
+### Push Interactions
+
+You might instead want to have a custom lookup of archive paths and
+media types. Let\'s say we start with this lookup, `archives`:
```
FeynmanZhou, line 108: Do we need to remove `\` in `Let\'s`?

vsoch: Yes I think so!

vsoch: okay I think I got most of them - let me know if you see one I missed! https://github.com/oras-project/oras-py/pull/41/commits/4a66955cbe84ebb0db6d030772f07e374f722574

docs/contributing.md
```diff
@@ -20,83 +20,5 @@ Please follow it in all your interactions with the project members and users.
 6. If necessary, update the README.md.
 7. The pull request will be reviewed by others, and the final merge must be done by an OWNER.
 
+For contributing to documentation, see our [Developer Guide](https://oras-project.github.io/oras-py/getting_started/developer-guide.html#documentation).
```
FeynmanZhou, line 23: Is it better to use a relative link?

docs/getting_started/developer-guide.md
```diff
@@ -0,0 +1,248 @@
...
+cd docs
+pip install -r requrements.txt
+
+# Build the docs into _build/html
+make html
+```
+
+### Preview Documentation
+
+After `make html` you can cd into `_build/html` and start a local web
```
FeynmanZhou, line 40: How about using` enter into` instead of `cd into` here?

docs/contributing.md
vsoch: Only if we need to move it. I think this originally started in the docs that render to the different domain, so I used the full url.

docs/getting_started/developer-guide.md
vsoch: I think (personally speaking) “enter into” reads more strangely than “cd into” because I’m used to the cd command.

vsoch: But if you feel strongly about it I can change it - I’m probably not the most representative user reading the docs!

docs/getting_started/installation.md
```diff
@@ -0,0 +1,55 @@
...
+Note that we have several extra modes for installation:
+
+``` console
+# interactions are done via the docker client instead of manual
+$ pip install oras[docker]
+
+# Install dependencies for linting and tests
+$ pip install oras[tests]
+
+# install everything
```
FeynmanZhou, line 29: For consistency, all sentences should be capitalized.

docs/getting_started/developer-guide.md
vsoch: okay this is changed.

docs/getting_started/installation.md
vsoch: All set!

FeynmanZhou (approved):

PR #42, opened 2022-08-15
adding support for manifest --annotation

vsoch:
This will provide similar functionality to https://github.com/oras-project/oras/pull/468/files.

I made one change - I allow an annotation file and `--annotation` given that a `$manifest` isn't provided in the file, and we could go even further to just not allow overlapping variable names. I think the tool should be as flexible as possible to different developer needs, but without having any unexpected results.


vsoch:
Seems that reviewing my PRs is not a priority, so I'll just take charge of the project and respond to issues that arise as needed.

Issue #49, opened 2022-10-26
Authentication against gitlab registry fails


magelisk:
Authentication against the Container registry in gitlab appears to fail. The cause is because the authentication expects the `service` parameter is expected to be a POST parameter, and not a header. 

A fix here is very easy in `authenticate_request` (https://github.com/oras-project/oras-py/blob/main/oras/provider.py#L691) 

```
        params = {}  # define further up

        h = oras.auth.parse_auth_header(authHeaderRaw)
        if h.service:
            params["service"] = h.service  # added
            headers.update(
                {
                    "Service": h.service,
                    "Accept": "application/json",
                    "User-Agent": "oras-py",
                }
            )
```

I know this works, I've patched in in my local branch and it works for both private gitlab container registries. However, I don't know if this may introduce compatibility issues with other container registries. I assume they would ignore unused params, but I don't know this for sure

# Output Details
When I run this below script I get the following console:
```
> registry = Registry(hostname="registry.gitlab.company.com:443")
> client = oras.client.OrasClient(registry=registry)
> res = client.pull(target="registry.gitlab.us.lmco.com:443/magelisk/artifact-testing/transfer-manifest.yaml:0.0.1", outdir="./TEMP")

authentication required
Issue with https://registry.gitlab.company.com:443/v2/magelisk/artifact-testing/transfer-manifest.yaml/manifests/0.0.1:
Unauthorized
```

The `originalResponse.headers` that goes into the `authenticate_request` function is:

```
{'Content-Type': 'application/json', 'Date': 'Wed, 26 Oct 2022 03:08:21 GMT', 'Docker-Distribution-Api-Version': 'registry/2.0', 'Server': 'nginx', 'Www-Authenticate': 'Bearer realm="https://gitlab.company.com/jwt/auth",service="container_registry",scope="repository:magelisk/artifact-testing/transfer-manifest.yaml:pull"', 'X-Content-Type-Options': 'nosniff', 'Content-Length': '197', 'Connection': 'keep-alive'}
```
The service of `container_registry` is definitely correct, but not in thne headers.

The way I actually figure out what was wrong was by comparing the ORAS CLI debug output. It gets the same 401, and you can see resulting request URL in last line with the param set

```
$ oras pull registry.gitlab.us.lmco.com:443/magelisk/artifact-testing/transfer-manifest.yaml:0.0.1 --debug

[...]

PR #50, opened 2022-10-26
Apply appropriate header for authentication against a gitlab registry

magelisk:
This adds a fix to set the `service=container_registry` to the request params rather than just headers. This is needed for gitlab registry authentication.

This is a simple fix for #49 , I don't know for certain if this may have negative impact for not gitlab registries.


vsoch:
You beat me to it!

so the Ora’s project requires signed commits, if you are able to do that. The other tweak is to bump the version in oras/version.py and add a note to the CHANGELOG.

magelisk:
I believe I got it all worked out. Signed and version+CHANGELOG updated

vsoch (approved):

PR #65, opened 2023-01-06
Update PACKAGE_URL with correct GH repo URL

yoanisgil:
Updating PACKAGE_URL as it was pointing to the wrong location, and hence the link [here](https://pypi.org/project/oras/)), is broken as it point to https://github.com/oras-project/oras-python


vsoch:
Thank you! It looks like you'll need to sign the DCO - click on the details to see how that works.

PR #68, opened 2023-02-03
Fix get tags

wolfv:
Even when asking for N tags, it doesn't mean that the (Github) registry returns that many tags. In my experience we need to use pagination, but I can't judge if this works for registries _other_ than ghcr.io (I would think so though).

This also changes the API and makes it conform to the types indicated (return `List[str]`), and not a response object. 

I tried to run the pre-commit linters but they are currently being pulled from a `local` repository? Wouldn't it be better to use the upstream ones? E.g.: https://github.com/mamba-org/mamba/blob/9de6782e02259e0d9050a1f74ac47383ff0d65f0/.pre-commit-config.yaml#L29-L43

### DIFF
--- a/oras/container.py
+++ b/oras/container.py
@@ -54,7 +54,9 @@ class Container:
     def upload_blob_url(self) -> str:
         return f"{self.registry}/v2/{self.api_prefix}/blobs/uploads/"
 
-    def tags_url(self, N=10_000) -> str:
+    def tags_url(self, N=10_000, query=None) -> str:
+        if query:
+            return f"{self.registry}/v2/{self.api_prefix}/tags/list?{query}"
         return f"{self.registry}/v2/{self.api_prefix}/tags/list?n={N}"
 
     def put_manifest_url(self) -> str:
--- a/oras/container.py
+++ b/oras/container.py
@@ -54,9 +54,7 @@ class Container:
     def upload_blob_url(self) -> str:
         return f"{self.registry}/v2/{self.api_prefix}/blobs/uploads/"
 
-    def tags_url(self, N=10_000, query=None) -> str:
-        if query:
-            return f"{self.registry}/v2/{self.api_prefix}/tags/list?{query}"
+    def tags_url(self, N=10_000) -> str:
         return f"{self.registry}/v2/{self.api_prefix}/tags/list?n={N}"
 
     def put_manifest_url(self) -> str:
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -4,6 +4,7 @@ __license__ = "Apache-2.0"
 
 import copy
 import os
+import urllib
 from typing import List, Optional, Tuple, Union
 
 import jsonschema
@@ -236,8 +237,34 @@ class Registry:
         :param N: number of tags
         :type N: int
         """
-        tags_url = f"{self.prefix}://{container.tags_url(N)}"  # type: ignore
-        return self.do_request(tags_url, "GET", headers=self.headers)
+        tags_url = f"{self.prefix}://{container.tags_url(N=N)}"  # type: ignore
+
+        tags: List[str] = []
+        has_next_link = True
+        # get all tags using the pagination
+        while len(tags) < N and has_next_link:
+            res = self.do_request(tags_url, "GET", headers=self.headers)
+
+            # raise before trying to get `json` value
+            res.raise_for_status()
+
+            if res.headers.get("Link"):
+                link = res.headers.get("Link")
+                # if we have a next link, that looks something like:
+                # </v2/channel-mirrors/conda-forge/linux-64/arrow-cpp/tags/list?last=2.0.0-py37h5a62f8a_45_cuda&n=100000000>; rel="next"
+                # we want to extract the url and get the rest of the tags
+                assert link.endswith('; rel="next"')
+                next_link = link[link.find("<") + 1 : link.find(">")]
+                query = urllib.parse.urlparse(next_link).query
+                tags_url = f"{self.prefix}://{container.tags_url(query=query)}"  # type: ignore
+            else:
+                has_next_link = False
+
+            # if the package does not exist, the response is an
+            # {"errors":[{"code":"NAME_UNKNOWN","message":"repository name not known to registry"}]}
+            tags += res.json().get("tags", [])
+
+        return tags
 
     @ensure_container
     def get_blob(
@@ -548,7 +575,6 @@ class Registry:
 
         # Upload files as blobs
         for blob in kwargs.get("files", []):
-
             # You can provide a blob + content type
             if ":" in str(blob):
                 blob, media_type = str(blob).split(":", 1)
@@ -809,7 +835,6 @@ class Registry:
         h = oras.auth.parse_auth_header(authHeaderRaw)
 
         if "Authorization" not in headers:
-
             # First try to request an anonymous token
             logger.debug("No Authorization, requesting anonymous token")
             if self.request_anonymous_token(h):
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -238,33 +238,69 @@ class Registry:
         :type N: int
         """
         tags_url = f"{self.prefix}://{container.tags_url(N=N)}"  # type: ignore
+        return self.do_paginated_request(tags_url, N=N, unwrap="tags")
 
-        tags: List[str] = []
+    def _get_response_links(self, response: requests.Response) -> dict:
+        """
+        Get a named or all response links.
+
+        response.links does not seem reliable to always parse, so we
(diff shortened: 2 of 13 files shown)


wolfv:
PS: this can be tested by asking for tags e.g. for this package: https://github.com/orgs/channel-mirrors/packages/container/package/conda-forge%2Flinux-aarch64%2Farrow-cpp

vsoch:
> Even when asking for N tags, it doesn't mean that the (Github) registry returns that many tags. In my experience we need to use pagination, but I can't judge if this works for registries other than ghcr.io (I would think so though).

I'm not sure, actually. I didn't think we'd hit the use case of wanting more than the default (is it 50?) but here we are! :laughing: 

> This also changes the API and makes it conform to the types indicated (return List[str]), and not a response object.

:+1: 

> I tried to run the pre-commit linters but they are currently being pulled from a local repository? 

There are different ways to go about this, but I like to install the deps directly to the same environment, so you need to do `pip install -r .github/dev-requirements.txt` first. There are instructions for that here: https://oras-project.github.io/oras-py/getting_started/developer-guide.html#code-linting and I'll show a screenshot below:

image

> Wouldn't it be better to use the upstream ones? E.g.: https://github.com/mamba-org/mamba/blob/9de6782e02259e0d9050a1f74ac47383ff0d65f0/.pre-commit-config.yaml#L29-L43

Maybe? "better" is relative I think - I might just like the better control / grouping with the same deps in my local environment.

vsoch:
@wolfv the DCO is unfortunately required by the upstream projects (under CNCF) and out of my control - let me know if you have any issues I can help with around that.

oras/provider.py
```diff
@@ -236,8 +237,31 @@ def get_tags(
         :param N: number of tags
         :type N: int
         """
-        tags_url = f"{self.prefix}://{container.tags_url(N)}"  # type: ignore
-        return self.do_request(tags_url, "GET", headers=self.headers)
+        tags_url = f"{self.prefix}://{container.tags_url(N=N)}"  # type: ignore
```
vsoch, line 240: We could probably verify this fairly quickly by arbitrarily pushing a ton of tags to the test registry in the actions and then retrieving them in this manner.

Is this pagination technique common, and if so, could it be generalized to a common function use by get tags? :thinking:

wolfv: I tried adding a test but the local registry returns all tags no matter what :) I found the docker specs though: https://docs.docker.com/registry/spec/api/#listing-image-tags where it's mentioning the pagination and some other discussion on github: 
https://github.com/orgs/community/discussions/25658 So that seems to be reasonably standard. 

I tried to find a configuration in the docker registry settings so that we could reproduce the pagination locally, but didn't find anything. 

We could test against the channel-mirrors package if that's something you'd want.

wolfv:
Hmm, looks like I might've received a newer black?

vsoch:
> Hmm, looks like I might've received a newer black?

Definitely possibly - we don't pin the version, and that is living a bit dangerously. It means if a newer version is grabbed in the CI, I will just update my local environment to use it. E.g., `pip install black==23.1.0`. Sorry :/

wolfv:
There are a couple unrelated changes now (some empty lines deleted). I don't care but if you do, I can revert them :)

oras/provider.py
wolfv: I added a small test that uses ghcr and that arrow-cpp package with > 1000 tags.

vsoch:
This is looking great! I'm in the middle of a work day and won't be able to review until after (when you are sleeping!) but I'll definitely have time this evening or this weekend to take a closer look, test stuffs locally, etc. With your permission, I can update the PR with any changes or do a PR to your branch if that is preferred. Thanks for doing this fix!

wolfv:
Thanks for your changes. Nice find on the `requests.Response` object, I also (obviously) didn't know about. I think it's much nicer to use it.

I made the pagination function a bit more generic by taking a callable to extract tags (or whatever). lmk what you think. To me, this looks good and works :)

oras/provider.py
```diff
@@ -233,11 +234,54 @@ def get_tags(
 
         :param container:  parsed container URI
         :type container: oras.container.Container or str
-        :param N: number of tags
+        :param N: number of tags, -1 for all
```
vsoch, line 237: -1 for all is something the endpoint honors?

oras/container.py
```diff
@@ -54,7 +54,9 @@ def get_blob_url(self, digest: str) -> str:
     def upload_blob_url(self) -> str:
         return f"{self.registry}/v2/{self.api_prefix}/blobs/uploads/"
 
-    def tags_url(self, N=10_000) -> str:
+    def tags_url(self, N=10_000, query=None) -> str:
```
vsoch, line 57: Do we still need to accept a query here?

vsoch (commented):
I really like these changes! If we have another case that doesn't warrant a callable we can make it optional, but I'm good leaving as is until that's proven. I also really like the response.links - I am glad you added it back.

The one question I have is about providing -1 for all - I think that's something we could honor here, but for the actual query I don't think I see it in the spec? https://github.com/opencontainers/distribution-spec/blob/067a0f5b0e256583bb9a088f72cba85ed043d1d2/spec.md?plain=1#L471-L513 So my suggestions (and let me know what you think!)

- have -1 be the default to retrieve all tags, but don't pass to the API endpoint.
- remove "query" from the tags_url property

After we discuss those things I think we should be good to go!

oras/container.py
wolfv: na, let's get rid of it! :)

oras/provider.py
wolfv: I completely missed that it would be sent to the endpoint haha. Apparently it works in the tests, but idk – should we make it `None` in the case of all tags and retrieve tags with N = "10k" or other random high number?

wolfv:
I changed it to use `optional[int]` for the `N` parameter. lmk what you think about that. I'll be afk for the rest of the night so feel free to take over and finish this PR if you want :)

Unfortunately `mypy` didn't accept somethign like


```
if not N:
   N = 10_000
```

which is why I had to add the temporary variable (mypy issue here: https://github.com/python/mypy/issues/4805

oras/provider.py
vsoch: I think as a user my expectation is to get back all tags - it would be annoying if I didn't! I think the easiest thing to do is just pass in the really high value for N, and if someone gives it None we would set to 0 and the distribution spec says a value of 0 should return an empty list. So something like:

```python
N = N or 0
```
I thought about some boolean like "retrieve_all" but then that could conflict with N and it gets too complicated, so I like just a high value of N. What do you think?

vsoch:
yep let me figure something out! And I'll get this merged / released asap after so we can test with the mirror.

vsoch (approved):

wolfv:
Uh oh now get all is limited to 10k tags :)

wolfv:
If I read the code correctly on mobile, we're now cutting the tags to N and N is set to 10k, even if we intend to retrieve all tags.

vsoch:
The logic hasn't changed - I think the bug we missed is that we don't know what to set the N parameter to in order to ask for all of them - should it be unset?

For our function - this should handle asking for more at least:

```
bool(len(new_tags) and (retrieve_all or len(tags) < N))
```
`retrieve_all` True would set `wants_more` to True. As for the N parameter in the query, you are right if that is going to set a limit we have a bug! I suppose it should be unset if we want all? :thinking:

vsoch:
okay looks like when it's unset it just returns a value of N as 0. I'm thinking if the value is -1 we can just set to None (so no N parameter required). I brought up a kubernetes cluster I need to work on for a bit but I'll do a suggested fix after that.

```
# When no query params are set:
response.links
Out[8]: 
{'next': {'url': '/v2/channel-mirrors/conda-forge/linux-aarch64/arrow-cpp/tags/list?last=1.0.1-py37h64c39f3_7_cpu&n=0',
  'rel': 'next'}}
```

vsoch:
This is a really good example of how mypy has influenced the actual design of the library to avoid having None as a default, and as a result has made this simple interaction more challenging. I'm not sure I'm a fan of these typing tools - beyond running the check they don't actually enforce anything!

wolfv:
Didn't you add a tags[:N] at the end of the function? That's what would cut everything beyond 10k for the retrieve all case if I read correctly

vsoch:
Oh that too. But if we ask the api (N) to go up to a certain point it’s not clear if it will stop paginating to. The fix is to allow the parameter to be None like we originally wanted I think and then not to try and set a default.

vsoch:
See https://github.com/oras-project/oras-py/pull/70

PR #70, opened 2023-02-04
bugfix that pagination sets upper limit when should not!

vsoch:
@wolfv see the changes here - I removed the typing because it was really just hurting us more than helping!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - bugfix that pagination sets upper limit of 10K (0.1.15)
  - pagination for tags (and general function for pagination) (0.1.14)
  - expose upload_blob function to be consistent (0.1.13)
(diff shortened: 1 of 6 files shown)


wolfv:
Hmm, I thought about pagination more and I think the `n` is actually OK because it just indicates how many to get per "page" (not in total). So maybe better to keep an arbitrarily high value to not use too many requests, but I haven't checked wether it makes a difference or not.

vsoch:
> Hmm, I thought about pagination more and I think the n is actually OK because it just indicates how many to get per "page" (not in total). So maybe better to keep an arbitrarily high value to not use too many requests, but I haven't checked wether it makes a difference or not.

I think N is typically how many results to get - it would be strange to only get a subset per page (but then to paginate!)

vsoch:
I updated the PR here to also include handling the case when tags is `null`. Let me know if we are ok to merge and release!

wolfv:
Yeah but 'n' would also be used to set the elements per page. But it also doesn't matter much to me as ping as it works :)

wolfv:
All good regarding merging and releasing btw!

vsoch:
It would be confusing for n to set elements per page, but have the endpoint be different (the total):

image

I've implemented / used a lot of APIs and (I think)? the parameter you have in mind is usually called `per_page` or similar.  

Let's get this out!

PR #71, opened 2023-02-05
fix bug when tags are `null`

wolfv:
unfortunately I discovered that `tags` can be `null`

### DIFF
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -249,7 +249,8 @@ class Registry:
             Determine if we should continue based on new tags and under limit.
             """
             json = response.json()
-            new_tags = json.get("tags", [])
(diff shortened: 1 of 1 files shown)


wolfv:
Also I was wondering if we should throw a different error type than `ValueError` when the HTTP request has a non-200 response? Maybe just throw whatever `requests` would throw?

vsoch:
That would work.

vsoch:
I fixed this bug in this branch so we can work off of one! https://github.com/oras-project/oras-py/pull/70/commits/8176eedefb061bd8ac14f24798fb49c0703fb6bb

vsoch:
This was fixed in #70 - thanks @wolfv !

PR #75, opened 2023-02-15
chore: add CODEOWNERS based on OWNERS.md

shizhMSFT:
Add [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) based on the [OWNERS.md](https://github.com/oras-project/oras-py/blob/main/OWNERS.md) file.

PR #77, opened 2023-02-16
fix: fix CODEOWNERS

shizhMSFT:
Having multiple entries with the same path results in that only the last entry takes effect.
This PR combines multiple entries with the same path into a single entry.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,2 @@
 # Derived from OWNERS.md
-* @vsoch
-* @jdolitsky
-* @stevelasker
+* @vsoch @jdolitsky @stevelasker


CODEOWNERS
```diff
@@ -1,4 +1,2 @@
 # Derived from OWNERS.md
-* @vsoch
-* @jdolitsky
-* @stevelasker
+* @vsoch @jdolitsky @stevelasker
```
vsoch, line 2: For reference:

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

Are you sure this file makes sense? E.g., we are going to be asking all three people for reviews every time there is a PR. I don't think either of the last two people have been involved with the development of the project so should we perhaps instead do it by file type, or bring in some contributors that have been active?

It was a bit awkward how you merged this file before, without talking to me, just for some feedback. Thanks for pinging this time around.

vsoch: If the two others on the list are ok with this ping, the changes are OK for me.

shizhMSFT: > I don't think either of the last two people have been involved with the development of the project so should we perhaps instead do it by file type, or bring in some contributors that have been active?

Are you suggesting that just keep you (@vsoch) in the `CODEOWNERS` file is good enough? If so, I will update this PR accordingly.

> It was a bit awkward how you merged this file before, without talking to me, just for some feedback. Thanks for pinging this time around.

Previous PR #75 was merged too fast than I thought. All owners are pinged in the [#oras slack channel](https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1676468542126809) for both PRs (#75 and #77).

shizhMSFT: > bring in some contributors that have been active

Yes, that's exactly one of purposes of this `CODEOWNERS` file.

vsoch: I’m good with whatever you’d like to do.

vsoch (approved):

Issue #78, opened 2023-03-02
Bug:  Session URL concatenates two url's into one bad one


jhlmco:
The following section of code incorrectly concatenates the URL into a bad session URL.

Example:
```
origional session_url: 'https://registry.gitlab.foo.com/v2/project/path/blobs/uploads/3e7bf2a9-xxxx-xxx-xxx-xxxxxxxx?_state=I8fblah'
prefix: 'https://registry.gitlab.foo.com:443'
```

Result:
```
session_url: 'https://registry.gitlab.foo.com:443https://registry.gitlab.foo.com/v2/project/path/blobs/uploads/3e7bf2a9-xxxx-xxx-xxx-xxxxxxxx?_state=I8fblah'
```

Because the registry has the port embedded in the URL the `session_url.startswith` concatenates it to the good session URL string.

https://github.com/oras-project/oras-py/blob/c817740aefba443817d2b994b243e54e189b8430/oras/provider.py#L496-L499

Expected actions:

Parse off the port from the URL when assigning the prefix to match.

vsoch:
It looks like GitLab returns a full session url that doesn't start with the prefix (the function assumes it either starts with the prefix or is a partial URL that needs it), so perhaps we should just check if it starts with `http`. Would you care to submit a PR to adjust this?

jhlmco:
> It looks like GitLab returns a full session url that doesn't start with the prefix (the function assumes it either starts with the prefix or is a partial URL that needs it), so perhaps we should just check if it starts with `http`. Would you care to submit a PR to adjust this?

@vsoch do you want something to check that it starts with that or do you want to check the protocol + hostname with something like urlparse (urllib)

vsoch:
I’ll leave to your judgment! The latter sounds more robust but if it’s overkill the simple approach works as well.

jhlmco:
@vsoch MR for review here: https://github.com/oras-project/oras-py/pull/79

kept it simple

PR #79, opened 2023-03-02
fix: Update the registry matching pattern

jhlmco:
Small fix, [closes issue 78](https://github.com/oras-project/oras-py/issues/78)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and **Merged pull requests**. Critical items to know are:
 
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
+## [0.1.17](https://github.com/oras-project/oras-py/tree/main) (0.1.17)
+  - patch fix to correct session url pattern, closes issue [78](https://github.com/oras-project/oras-py/issues/78)
+
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
  - add support for tag deletion and retry decorators (0.1.16)
  - bugfix that pagination sets upper limit of 10K (0.1.15)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,6 @@ and **Merged pull requests**. Critical items to know are:
 
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
-## [0.1.17](https://github.com/oras-project/oras-py/tree/main) (0.1.17)
-  - patch fix to correct session url pattern, closes issue [78](https://github.com/oras-project/oras-py/issues/78)
-
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
   - patch fix to correct session url pattern, closes issue [78](https://github.com/oras-project/oras-py/issues/78) (0.1.17)
   - add support for tag deletion and retry decorators (0.1.16)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ The versions coincide with releases on pip. Only major versions will be released
   - patch fix to correct session url pattern, closes issue [78](https://github.com/oras-project/oras-py/issues/78)
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
- - add support for tag deletion and retry decorators (0.1.16)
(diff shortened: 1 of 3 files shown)


vsoch:
Awesome! You'll need to sign the DCO to contribute (not my rule, inherited because oras is a CNCF project  I think) and then:

 - bump the version in oras/version.py
 - update the CHANGELOG.md with a note and the new version

jhlmco:
> Awesome! You'll need to sign the DCO to contribute (not my rule, inherited because oras is a CNCF project I think) and then:
> 
> * bump the version in oras/version.py
> * update the CHANGELOG.md with a note and the new version

I'm actually not sure why it's saying that.  Under commits my last two commit messages have the DCO signature.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and **Merged pull requests**. Critical items to know are:
 
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
+## [0.1.17](https://github.com/oras-project/oras-py/tree/main) (0.1.17)
+  - patch fix to correct session url pattern, closes issue [78](https://github.com/oras-project/oras-py/issues/78)
+
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
  - add support for tag deletion and retry decorators (0.1.16)
```
vsoch, line 20: This should follow the previous convention (not a new section) but rather:

```suggestion
  - patch fix to correct session url pattern, closes issue [78](https://github.com/oras-project/oras-py/issues/78) (0.1.17)
  - add support for tag deletion and retry decorators (0.1.16)
```

oras/provider.py
```diff
@@ -492,8 +492,10 @@ def _get_location(
         if not session_url:
             return session_url
 
-        # Some registries do not return the full registry hostname
-        prefix = f"{self.prefix}://{container.registry}"
+        # Some registries do not return the full registry hostname.  Check that 
+        # the url starts with a protocol scheme, change tracked with:
```
vsoch, line 496: We won't want to replace prefix here - for most cases it will be something meaningful. Rather, I think we want:

```python
prefix = f"{self.prefix}://{container.registry}"

if not session_url.startswith('http'):
    # same
```

jhlmco: @vsoch prefix isn't used anywhere else in that method.  So retaining the variable doesn't make sense if we're going to just use the static string

jhlmco: Wait that would bread the original fix.  I'll update and push.

vsoch: So the endpoint can return a partial url, meaning one that is just the original self.prefix (http or https) and then the container registry URI. If you just define prefix to be http, it's going to miss the entire registry URL and just return a truncated url.

jhlmco: Good catch, should be fixed

CHANGELOG.md
```diff
@@ -13,8 +13,12 @@ and **Merged pull requests**. Critical items to know are:
 
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
+## [0.1.17](https://github.com/oras-project/oras-py/tree/main) (0.1.17)
```
vsoch, line 16: Did you forget to remove this extra section?

jhlmco: fixed.

vsoch:
@jhlmco you likely have a different version of black - make sure you are using `23.1.0`

jhlmco:
> @jhlmco you likely have a different version of black - make sure you are using `23.1.0`

Yeah, I created a virtual environment, installed everything, wired in pre-commit, and ran it.  I get a ton of errors for mypi but nothing to do with the 2 line change I made.

vsoch:
Would you like some help?

jhlmco:
@vsoch if it's a quick formatting fix that'd be great!

vsoch:
This looks great! I'll try to remove my sign when I squash and merge so the commit credit is fully yours. Thanks for your patience with the linting!

jhlmco:
@vsoch Awesome!  Glad to help!  Looking to leverage it in another opensource tool (shameless plug [hoppr](https://hoppr.dev)) as an option for bundling artifacts.

vsoch:
ohh that's super adorable!! I'm sold :heart_eyes: Please ping me if/when you have something to test.

jhlmco:
Totally!

Issue #90, opened 2023-07-04
OrasClient.push fails on Windows as empty Manifest Config uses /dev/null which is Linux specific


sunnycarter:
My code runs fine on Linux.  On Windows it fails with an error.

Here's my code:
```
        client = OrasClient(hostname=acr_url)
        
        client.login(
            username=self._manifest_credentials["username"],
            password=self._manifest_credentials["acr_token"],
        )
        target = (
                f"{client.remote.hostname.replace('https://', '')}/testartifact:1.0.0"
            )
        client.push(`
                files=["testartifact.json"],
                target=target
            )
```
```
 File "<>\oras\client.py", line 131, in push
    return self.remote.push(*args, **kwargs)
  File "<>\oras\provider.py", line 718, in push
    response = self.upload_blob(config_file, container, conf)
  File "<>\oras\provider.py", line 217, in upload_blob
    response = self.put_upload(blob, container, layer)
  File "<>\oras\provider.py", line 474, in put_upload
    with open(blob, "rb") as fd:
OSError: [Errno 22] Invalid argument: 'C:\\dev\\null'
```
I have made a fix for this and will submit a PR.  The fix is in oci.py ManifestConfig:
```
        if platform.system() == "Windows":
            path = "nul"
        else:
            path = "/dev/null"
```

sunnycarter:
https://github.com/oras-project/oras-py/pull/91 submitted to fix this

sunnycarter:
Fixed in 0.1.18

PR #91, opened 2023-07-04
Fix issue 90 - empty manifest config support on Windows

sunnycarter:
Fix for https://github.com/oras-project/oras-py/issues/90

Tested with repro on Windows and Linux.

I'm not quite sure what to do with the CHANGELOG.md I'm afraid so please give guidance.

### DIFF
--- a/oras/oci.py
+++ b/oras/oci.py
@@ -4,6 +4,7 @@ __license__ = "Apache-2.0"
 
 import copy
 import os
+import platform
 from typing import Dict, Optional, Tuple
 
 import jsonschema
@@ -128,7 +129,10 @@ def ManifestConfig(
     """
     # Create an empty config if we don't have one
     if not path or not os.path.exists(path):
-        path = "/dev/null"
+        if platform.system() == "Windows":
+            path = "nul"
+        else:
+            path = "/dev/null"
         conf = {
             "mediaType": media_type or oras.defaults.unknown_config_media_type,
             "size": 0,
--- a/oras/oci.py
+++ b/oras/oci.py
@@ -4,7 +4,6 @@ __license__ = "Apache-2.0"
 
 import copy
 import os
-import platform
 from typing import Dict, Optional, Tuple
 
 import jsonschema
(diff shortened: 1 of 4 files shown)


oras/oci.py
```diff
@@ -128,7 +129,10 @@ def ManifestConfig(
     """
     # Create an empty config if we don't have one
     if not path or not os.path.exists(path):
-        path = "/dev/null"
+        if platform.system() == "Windows":
+            path = "nul"
+        else:
+            path = "/dev/null"
```
vsoch, line 135: Here you can do:

```python
path = "nul" if platform.system() == "Windows" else "/dev/null" 
```

oras/provider.py
```diff
@@ -430,7 +431,10 @@ def download_blob(
         # Allow an empty layer to fail and return /dev/null
         except Exception as e:
             if digest == oras.defaults.blank_hash:
-                return "/dev/null"
+                if platform.system() == "Windows":
```
vsoch, line 434: If we are doing this twice - maybe write a util function instead?

oras/oci.py
sunnycarter: I found a better way to fix with os.devnull

oras/provider.py
sunnycarter: Found that os.devnull exists

sunnycarter:
Thanks for the review, the nod to write a utility function led me to discover there's an inbuilt one

vsoch:
> Thanks for the review, the nod to write a utility function led me to discover there's an inbuilt one

Nice! And good intuition to look for that too!

For the DCO you can click on "Details" to learn more about that - if I remember it's just an electronic page you need to agree to and then you are good. It would only be done once! The ORAS family of projects is under the CNCF umbrella so we follow these policies that are enforced there.

vsoch:
> I'm not quite sure what to do with the CHANGELOG.md I'm afraid so please give guidance.

For this, you'll want to increment the version to 0.1.18 here: https://github.com/oras-project/oras-py/blob/209c9b98043a00d1b04789cc2967ca7021dc5b2e/oras/version.py#L5

And then make a note (with that corresponding version) on a newline above 0.1.17 here  https://github.com/oras-project/oras-py/blob/main/CHANGELOG.md

sunnycarter:
Thanks for looking at this, I've now signed all commits to fulfil the DCO, and pushed a new commit with a patch version bump and changelog update.

vsoch (approved):

Issue #95, opened 2023-07-15
[Feature] Skip tls verifikation


mariusbertram:
Hello,
it would be nice feature to be able to skip tls verification for the Oras.Client.
Like on the oras cli.

vsoch:
Did you try the insecure parameter? https://github.com/oras-project/oras-py/blob/8c1f7cc64a97e0580024b642c6eefeb07968799b/oras/client.py#L32

mariusbertram:
Yes I tried the parameter. But I get this Error on MacOS with Python 3.11.4 and in an alpine:python3.11 image.
`raise SSLError(e, request=request)
requests.exceptions.SSLError: HTTPSConnectionPool(host='192.168.178.43', port=443)`

mariusbertram:
I think it's a bug. The insecure parameter is not pass through to oras.provide and do_request() 
https://github.com/oras-project/oras-py/blob/8c1f7cc64a97e0580024b642c6eefeb07968799b/oras/provider.py#L810

mariusbertram:
I made a pull request with an fix #96

vsoch:
Thank you @mariusbertram ! It should be released to pypi shortly, version 0.1.2.

mariusbertram:
The Version 0.1.2 is on pypi but version 0.1.2 is smaller than 0.1.19. I think this version schould be 0.1.20

vsoch:
Good call - let me fix that up.

PR #96, opened 2023-07-15
Add insecure option to requests

mariusbertram:
I fixed the found Bug in #95 .
Now all requests have the option to skip tls verification.

### DIFF
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -846,7 +846,7 @@ class Registry:
             json=json,
             headers=headers,
             stream=stream,
-            verify=not self._insecure
+            verify=not self._insecure,
         )
 
         # A 401 response is a request for authentication
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -251,7 +251,6 @@ class Registry:
             head_url,
             "HEAD",
             headers={"Accept": "application/vnd.oci.image.manifest.v1+json"},
-            insecure=self._insecure
         )
         if response.status_code == 404:
             logger.error(f"Cannot find tag {container}:{tag}")
@@ -262,7 +261,7 @@ class Registry:
             raise RuntimeError("Expected to find Docker-Content-Digest header.")
 
         delete_url = f"{self.prefix}://{container.manifest_url(digest)}"  # type: ignore
-        response = self.do_request(delete_url, "DELETE", insecure=self._insecure)
+        response = self.do_request(delete_url, "DELETE")
         if response.status_code != 202:
             raise RuntimeError("Delete was not successful: {response.json()}")
         return True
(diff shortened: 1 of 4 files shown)


oras/provider.py
```diff
@@ -348,7 +353,7 @@ def get_blob(
         """
         method = "GET" if not head else "HEAD"
         blob_url = f"{self.prefix}://{container.get_blob_url(digest)}"  # type: ignore
-        return self.do_request(blob_url, method, headers=self.headers, stream=stream)
+        return self.do_request(blob_url, method, headers=self.headers, stream=stream, insecure=self._insecure )
```
vsoch, line 356: Thank you for opening this! 

Perhaps we don't need to pass self._insecure into a function that already is on the class - we could just use it in the function without doing that?

vsoch (commented):
Thank you for opening this! Will you have any issues signing the DCO? This is a CNCF project so it's standard policy.

oras/provider.py
mariusbertram: You are right. I made a new commit where the function do_request gets the insecure parameter from the class

oras/provider.py
```diff
@@ -348,7 +352,7 @@ def get_blob(
         """
         method = "GET" if not head else "HEAD"
         blob_url = f"{self.prefix}://{container.get_blob_url(digest)}"  # type: ignore
-        return self.do_request(blob_url, method, headers=self.headers, stream=stream)
+        return self.do_request(blob_url, method, headers=self.headers, stream=stream )
```
vsoch, line 355:
```suggestion
        return self.do_request(blob_url, method, headers=self.headers, stream=stream)
```

oras/provider.py
```diff
@@ -842,6 +846,7 @@ def do_request(
             json=json,
             headers=headers,
             stream=stream,
+            verify=not self._insecure
```
vsoch, line 849: Perfecto! This is exactly what I had in mind.

oras/provider.py
vsoch: This (and the trailing comma you edited) will likely fail the formatting check. I will approve this to run but we will still need the DCO signed (let me know if you have questions about that).

mariusbertram: I signed the commits with my ssh key. Or does it needed with with gpg ? I will fix the formarting.

mariusbertram: Formarting is back to origin :D

vsoch: It's something called a DCO that you need to sign for CNCF projects - click the link in the failed step below, e.g., https://github.com/oras-project/oras-py/pull/96/checks?check_run_id=15068749650

mariusbertram: Ahh I didn't know that. Its done

vsoch:
ah - there was a bump to black that introduced some error. I can fix this for you.

PR #98, opened 2023-07-16
Distinguish between insecure and tis-verify

mariusbertram:
The insecure option for oras.registry sets the transport protocol. The default behaviour of other projects which interact with registries (ores-cli, podman,....),  differenciate between transport protocol and trust for connections to registries.
To avoid breaking changes, the parameter `insecure` sets the transport protocol ans the new parameter `tls_verify` sets verification on CA Trust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
-- add tls_verify to provider class for optional disable tls verification (0.1.22)
+ - add tls_verify to provider class for optional disable tls verification (0.1.22)
  - Allow to pull exactly to PWD (0.1.21)
  - Ensure insecure is passed to provider class (0.1.20)
  - patch fix for blob upload Windows, closes issue [93](https://github.com/oras-project/oras-py/issues/93) (0.1.19)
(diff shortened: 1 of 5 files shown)


vsoch (commented):
Should this be exposed from the oras/client.py as well? And don't forget to add a note to the CHANGELOG.md and bump the version in oras/version.py

mariusbertram:
Yes, there is a different between a registry listening with plain http  and listening on https with https but not a valid Crt.
I don't append the port for plain http  the client try to connect to port 80 and on https on port 443. Best practice configuration for the registry/loadbalancer/webserver is to redirect from port 80 to 443 with a rewrite from http to https.
But if you only expose Port 443 and you want to connect to an insecure registry you will not be able to connect with the current behaviour.  I will extend the changelog and bump the Version.

vsoch:
I think you missed my question about the client - could we please expose the parameter here along with insecure? https://github.com/oras-project/oras-py/blob/0f331fb11e5eb852fd8653009f650ede7094aa91/oras/client.py#L32

mariusbertram:
The docker client will skip the tis verification if tis_verify is set. On Plain http registries tis_verify has no effect.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+- add tls_verify to provider class for optional disable tls verification (0.1.22)
  - Allow to pull exactly to PWD (0.1.21)
```
vsoch, line 18:
```suggestion
 - add tls_verify to provider class for optional disable tls verification (0.1.22)
 - Allow to pull exactly to PWD (0.1.21)
```

vsoch (approved):

Issue #100, opened 2023-07-25
Directory mismatch on pull


bootstrap314:
When I pull an image, the directory specified by "outdir" is repeated. This causes the extracted files to be saved to the wrong directory:

For example, with the following code, this is my result:

```import oras.client
import os

from oras.logger import logger, setup_logger

client = oras.client.OrasClient()
print(client.version())

f = open("secret.json", "r")
password = f.read()

image = <image name:tag>

def pull_to_oci_layout(username, password, outdir, reference):
    try:
        hostname = reference.split("/")[0]
        result = client.login("_json_key", password, hostname=hostname)
        print(result)
        client.pull(
            allowed_media_type=[],
            overwrite=True,
            outdir="tmp",
            target=reference,
            debug=True,
        )
    except Exception as e:
        logger.exit(str(e))


pull_to_oci_layout(
    "_json_key",
    password,
    "tmp",
    image,
)
```

This yields the directory structure:
```
tmp
└── tmp
    ├── Cargo.lock
    ├── Cargo.toml
    ├── config.json
    ├── loader.wasm
    └── src
        ├── foundation.rs
        ├── foundation_callbacks.rs
        ├── gas.rs
        ├── getters.rs
        ├── internal.rs
        ├── lib.rs
        ├── owner.rs
        ├── owner_callbacks.rs
        ├── tests
        │   └── test_utils.rs
        └── types.rs
```

I've been trying to figure out why this is occurring but haven't figured it out yet. Has anyone else seen this?

vsoch:
The logic you’d want to debug starts here: https://github.com/oras-project/oras-py/blob/db32c9828c8914eca40ba1cc86c7974b14b129c6/oras/provider.py#L774

saketjajoo:
@vsoch, I am facing the same problem that the `outdir` is being repeated twice. Also, path aliases (e.g. `~` in mac = `$HOME` dir) do not get resolved. Example:
* outdir = `~/data` yields the complete path to be `$PWD/~/data/~/data/`
* outdir = `./data` yields the complete path to be `$PWD/data/data`

I believe I have a fix for this. Can I pick up this issue and propose a PR to fix it?

vsoch:
Yes that would be excellent - please go ahead.

saketjajoo:
PR to fix this: https://github.com/oras-project/oras-py/pull/106

saketjajoo:
This issue can be marked as fixed/completed!

vsoch:
[...]

PR #106, opened 2023-09-21
Removing the additional duplicate sub-directory from sanitize_path()

saketjajoo:
The existing implementation adds an additional sub-directory with the same name when an image is pulled and is stored explicitly to a custom output directory. This happens because the "test_path" variable [appends](https://github.com/oras-project/oras-py/blob/main/oras/utils/fileio.py#L50) "path" to "base_dir" wherein the ["path" variable itself is base_dir + filename](https://github.com/oras-project/oras-py/blob/main/oras/provider.py#L794). This commit fixes this problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - eliminate the additional subdirectory creation while pulling an image to a custom output directory (0.1.24)
   - patch fix for pulling artifacts by digest (0.1.23)
     - patch fix to reject cookies as this could trigger registries into handling the lib as a web client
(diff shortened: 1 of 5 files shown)


saketjajoo:
This PR is a fix to https://github.com/oras-project/oras-py/issues/100.

vsoch:
A few requests as tests are running (and I'm switching back and forth to other work):

 - Let's please add tests that are outside of home, and that include relative and absolute paths
 - Please bump the version in oras/version.py
 - Make a corresponding note with the version in CHANGELOG.md

saketjajoo:
It seems that Python's [black library](https://pypi.org/project/black/) recently made a [change](https://github.com/psf/black/commit/cf4cc2981900565ab931aada176abf08a1f5782d) to validate the data type of the "exclude" key in the pyproject.toml file. Now the value for "exclude" must be a [single-quoted string](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format) (for regex matches). I've added a second commit to this PR to address this.

vsoch (approved):

PR #114, opened 2023-11-10
bug with .tar.gz upload - name needs .tar.gz in it

vsoch:
There is a bug with the title that, given it does not have .tar.gz with the oras-go client, it will not be extracted as such.


MichaelKopfMkf:
I think this is not the desired behaviour. 
The oras cli does not publish directories with a tar.gz file ending in the title.

I think it uses the layers annotation `io.deis.oras.content.unpack` to decide whether to unpack a layer or not. 
the title is then used as target location for the unpacked tar.

I suggested to make oras-py also use this annotation here:
https://github.com/oras-project/oras-py/pull/170

Issue #112, opened 2023-10-26
Error due to default value of manifest_config


linshokaku:
If manifest_config is not specified when pushing by oras-py, the implementation currently treats manifest_config as a completely empty string.
https://github.com/oras-project/oras-py/blob/02c5bdacb54b5d7d6c152fcdbc0e11cf69b4900f/oras/oci.py#L130-L136

I was using google artifact registry and confirmed that push fails if manifest_config is an empty string.

The current workaround (or maybe this is the right way) is to create a file with only "{}" and read that file to successfully execute the push.

If manifest_config was not specified, why not change it so that the contents of manifest_config are treated as "{}"?

# code

```python
import oras.client
import oras.logger
from pathlib import Path
from tempfile import TemporaryDirectory
import os
import json

oras.logger.setup_logger(debug=True, quiet=False)

pwd = Path(__file__).parent.resolve()

client = oras.client.OrasClient()
client.login(username="oauth2accesstoken", password="password")

with TemporaryDirectory() as tmp:
    config_path = os.path.join(tmp, "config.json")
    with open(config_path, "w") as f:
        json.dump({}, f)
    res = client.push(files=[str(pwd / "hoge.txt")], target="asia-northeast1-docker.pkg.dev/example-repository/tmp/oras-test:v2", manifest_config=config_path)
print("sucess push with config file")

res = client.push(files=[str(pwd / "hoge.txt")], target="asia-northeast1-docker.pkg.dev/example-repository/tmp/oras-test:v2")
print("success push")
```

# log

```
Preparing layer {'mediaType': 'application/vnd.oci.image.layer.v1.tar', 'size': 13, 'digest': 'sha256:8f8ad85c91228f6b241b95ecca626a4e9701d7c072ca0c6c001677d797a4af02', 'annotations': {'org.opencontainers.image.title': 'hoge.txt'}}
Service: asia-northeast1-docker.pkg.dev
Scope: repository:example-repository/tmp/oras-test:pull,push
Preparing config {'mediaType': 'application/vnd.unknown.config.v1+json', 'size': 2, 'digest': 'sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a'}
Service: asia-northeast1-docker.pkg.dev
Scope: repository:example-repository/tmp/oras-test:pull,push
Service: asia-northeast1-docker.pkg.dev
Scope: repository:example-repository/tmp/oras-test:pull,push
[...]

PR #115, opened 2023-11-13
Allow uploading when  `config_manifest` is not specified in the google artifact registry

linshokaku:
#112 

The issue causing errors in Google Artifact Registry when no config_manifest was specified has been resolved.

I could not reproduce this problem with the test method shown in the [Developer's Guide](https://oras-project.github.io/oras-py/getting_started/developer-guide.html#documentation), but I have confirmed that it is resolved in my own environment.

### DIFF
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -7,6 +7,7 @@ import os
 import urllib
 from dataclasses import asdict, dataclass
 from http.cookiejar import DefaultCookiePolicy
+from tempfile import TemporaryDirectory
 from typing import Callable, List, Optional, Tuple, Union
 
 import jsonschema
@@ -747,7 +748,15 @@ class Registry:
 
         # Config is just another layer blob!
         logger.debug(f"Preparing config {conf}")
-        response = self.upload_blob(config_file, container, conf)
+        if config_file is None:
+            with TemporaryDirectory() as tmp:
+                config_file = os.path.join(tmp, "config.json")
+                with open(config_file, "w") as f:
+                    f.write("{}")
+                response = self.upload_blob(config_file, container, conf)
+        else:
+            response = self.upload_blob(config_file, container, conf)
+
         self._check_200_response(response)
 
         # Final upload of the manifest
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -7,7 +7,6 @@ import os
 import urllib
 from dataclasses import asdict, dataclass
 from http.cookiejar import DefaultCookiePolicy
-from tempfile import TemporaryDirectory
 from typing import Callable, List, Optional, Tuple, Union
 
 import jsonschema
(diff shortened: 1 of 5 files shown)


oras/provider.py
```diff
@@ -747,7 +748,15 @@ def push(self, *args, **kwargs) -> requests.Response:
 
         # Config is just another layer blob!
         logger.debug(f"Preparing config {conf}")
-        response = self.upload_blob(config_file, container, conf)
+        if config_file is None:
+            with TemporaryDirectory() as tmp:
+                config_file = os.path.join(tmp, "config.json")
+                with open(config_file, "w") as f:
+                    f.write("{}")
+                response = self.upload_blob(config_file, container, conf)
```
vsoch, line 756: This response block happens either way, so why not just have the "if none do x" and then remove the else, and no matter what call response = xxxx?

linshokaku: If no file is specified, a temporary file must be created before sending, and then the resources provided to create the temporary file must be released.

Therefore, if no file is specified, it is necessary to execute `upload_blob()` in the scope under `with TemporaryDirectory() as tmp:`.

If a single call to `upload_blos` code is desired, the code would be such that it would require execution of `TemporaryDirectory()` even if the file exists.
```python
with TemporaryDirectory() as tmp:
    if config_file is None:
        config_file = os.path.join(tmp, "config.json")
        with open(config_file, "w") as f:
            f.write("{}")
    response = self.upload_blob(config_file, container, conf)
```

Is this the correct modification?

vsoch: I think the "with" context is adding the extra else here - and I see your point about needing to clean up after (and it not working with the context). Would it be too simple to use the already provided functions https://github.com/oras-project/oras-py/blob/02c5bdacb54b5d7d6c152fcdbc0e11cf69b4900f/oras/utils/fileio.py#L168 to create a directory or file, and then just clean it up after? And remove the redundant logic using the with context? :thinking:

linshokaku: Fixed using `oras.utils.fileio.get_tmpfile()` .

oras/provider.py
```diff
@@ -747,7 +747,12 @@ def push(self, *args, **kwargs) -> requests.Response:
 
         # Config is just another layer blob!
         logger.debug(f"Preparing config {conf}")
+        if config_file is None:
+            config_file = oras.utils.get_tmpfile(suffix=".json")
+            with open(config_file, "w") as f:
```
vsoch, line 752: You can use oras.utils.write_file here: https://github.com/oras-project/oras-py/blob/02c5bdacb54b5d7d6c152fcdbc0e11cf69b4900f/oras/utils/fileio.py#L235 should just need to do:

```python
oras.utils.write_file(config_file, "{}")
```

vsoch: And I guess we are OK here not cleaning up?

linshokaku: * Fixing the use of write_file
* Delete temporary files

Two corrections were made.

vsoch (approved):

vsoch:
Thanks!

PR #123, opened 2024-01-29
Add option to refresh (current behaviour) or keep previous headers when fetching manifests

kavish-p:
Related PR: #121 

A new optional parameter `refresh_headers` has been added to the `pull` and `get_manifest` functions.
When no value is provided, `refresh_headers` is set to True. This in in line with the current behaviour.

Setting `refresh_headers` to False carries forward previously set headers to the `get_manifest` call.


kavish-p:
hi @vsoch , sorry, I messed up my local repository so I thought it would be easier to create a new PR.

Regarding your comment from https://github.com/oras-project/oras-py/pull/121#issuecomment-1913905195: When refresh_headers is not defined (None), I believe the default value should be True. Then previously set headers are ignored, which is the current behaviour.

vsoch (approved):
This looks fantastic! Thank you for cleaning up the git history - it's much easier to review. It looks like this can go in without changing any default behavior and address your need, so let's ship it! I'll trigger a release shortly after.

Thank you!

oras/provider.py
```diff
@@ -788,9 +792,12 @@ def pull(self, *args, **kwargs) -> List[str]:
         :type target: str
         """
         allowed_media_type = kwargs.get("allowed_media_type")
+        refresh_headers = kwargs.get("refresh_headers")
+        if refresh_headers is None:
+            refresh_headers = True
         container = self.get_container(kwargs["target"])
         self.load_configs(container, configs=kwargs.get("config_path"))
-        manifest = self.get_manifest(container, allowed_media_type)
+        manifest = self.get_manifest(container, allowed_media_type, refresh_headers)
```
prabhu, line 800: This is a breaking change that has broken depscan. It shouldn't have been released as a patch.

vsoch: Would you care to elaborate on that? This change adds no new dependencies, and (without specifying anything) the default behavior is the same as before.

vsoch: Ah I see, you maintain a project, depscan https://github.com/owasp-dep-scan/dep-scan/actions/runs/7695237445/job/20967667418#step:7:44 and you have a custom class that reimplements the function https://github.com/owasp-dep-scan/dep-scan/blob/db4988ca72e6a6c28befec378191f62b2ab340bc/depscan/lib/orasclient.py#L25. You just need to add an additional argument there.

We can't always anticipate how a user of the library is going to customize a class for their needs. For your implementation it looks like it's a [mirror copy](https://github.com/oras-project/oras-py/blob/6c9fd25cddbdcea0bb5929a07dfb5378edeb786c/oras/provider.py#L839) but you removed validation. My suggestion would be to not re-implement the function but request a parameter that might more easily turn off the validation if you don't want it.

PR #131, opened 2024-05-02
feat: simplify how subjects are created from existing manifest

tumido:
Hi folks 👋 

I've noticed the Python Oras client library has kind of rudimentary support for [attached artifacts](https://oras.land/docs/concepts/reftypes/). I'd like to address a couple of UX caveats in several PRs. Please let me know if it's heading in the right direction or if you have a different perspective.

**This PR addresses a difficulty when creating a Subject reference.**

As of today, an `oras-py` user needs to know how the subject spec is constructed, and how to calculate size and digest. This makes it quite hard to come up with a usable structure that can be used when pushing an artifact. I've addressed this by adding a `from_manifest` class method to the `Subject` data class. There may be a nicer, cleaner solution than a class method, let me know what you think.

### DIFF
--- a/docs/getting_started/user-guide.md
+++ b/docs/getting_started/user-guide.md
@@ -489,6 +489,28 @@ def push(uri, root):
 
 </details>
 
+<details>
+
+<summary>Example of basic artifact attachment</summary>
+
+We are assuming an `derived-artifact.txt` in the present working directory and that there's already a `localhost:5000/dinosaur/artifact:v1` artifact present in the registry. Here is an example of how to [attach](https://oras.land/docs/concepts/reftypes/) a derived artifact to the existing artifact.
+
+```python
+import oras.client
+import oras.oci
+
+client = oras.client.OrasClient(insecure=True)
+
+manifest = client.remote.get_manifest("localhost:5000/dinosaur/artifact:v1")
+subject = oras.oci.Subject.from_manifest(manifest)
+
+client.push(files=["derived-artifact.txt"], target="localhost:5000/dinosaur/artifact:v1-derived", subject=subject)
+Successfully pushed localhost:5000/dinosaur/artifact:v1-derived
+Out[4]: <Response [201]>
+```
+
+</details>
+
 The above examples are just a start! See our [examples](https://github.com/oras-project/oras-py/tree/main/examples)
 folder alongside the repository for more code examples and clients. If you would like help
 for an example, or to contribute an example, [you know what to do](https://github.com/oras-project/oras-py/issues)!
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -6,7 +6,7 @@ import copy
 import os
 import urllib
 from contextlib import contextmanager, nullcontext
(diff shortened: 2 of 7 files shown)


docs/getting_started/user-guide.md
```diff
@@ -489,6 +489,28 @@ def push(uri, root):
...
+
+We are assuming an `derived-artifact.txt` in the present working directory and that there's already a `localhost:5000/dinosaur/artifact:v1` artifact present in the registry. Here is an example of how to [attach](https://oras.land/docs/concepts/reftypes/) a derived artifact to the existing artifact.
+
+```python
+import oras.client
+import oras.provider
+
+client = oras.client.OrasClient(insecure=True)
+
+manifest = client.remote.get_manifest(f"localhost:5000/dinosaur/artifact:v1")
```
vsoch, line 504:
```suggestion
manifest = client.remote.get_manifest("localhost:5000/dinosaur/artifact:v1")
```

oras/provider.py
```diff
@@ -41,6 +44,23 @@ class Subject:
     digest: str
     size: int
 
+    @classmethod
```
vsoch, line 47: I like this design! I think (now that it's more substantial) we might consider moving Subject into `oras.oci` so it's clearly organized as an OCI type. What do you think?

vsoch (commented):
I like how this works! The reason the support is simple is because the interaction is more complex than seems at face value. If we want to fully support a client doing this interaction, the full thing is described here: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests-with-subject Let me know your thoughts.

oras/provider.py
tumido: That makes perfect sense to me. I'll move it. 👍

tumido:
> If we want to fully support a client doing this interaction, the full thing is described here: [opencontainers/distribution-spec@main/spec.md#pushing-manifests-with-subject](https://github.com/opencontainers/distribution-spec/blob/main/spec.md?rgh-link-date=2024-05-02T13%3A41%3A50Z#pushing-manifests-with-subject) Let me know your thoughts.

Well... I kinda don't like when things are half-done. 🙂 So.. I'll look into it and try to improve the push logic. (I may not have much time in the next two weeks though...)

Do you think it would make sense to split the `push` logic into `push` and `attach` methods, similarly to what `oras` CLI does? https://oras.land/docs/commands/oras_attach

vsoch:
> Do you think it would make sense to split the push logic into push and attach methods, similarly to what oras CLI does? https://oras.land/docs/commands/oras_attach

Since we don't have a command line client, let's start with what you have now - a push that takes a subject. If "attach" comes down to just doing that command again, we don't need attach. But if there is another use case not covered, we can consider adding it.

vsoch:
You'll want to finish up by:

- Bump the version in oras/version.py
- Adding a line to the CHANGELOG.md with the change and new version

tumido:
Done, thank you. 🙂

PR #132, opened 2024-05-02
feat: pull recursively to include subject references

tumido:
Hi folks 👋 

I've noticed the Python Oras client library has kind of rudimentary support for [attached artifacts](https://oras.land/docs/concepts/reftypes/). I'd like to address a couple of UX caveats in several PRs. Please let me know if it's heading in the right direction or if you have a different perspective.

**Provide means to pull referred artifacts similar to `oras pull --include-subject`:**

While `oras-py` supports including a subject when pushing a manifest, it does not support pulling the referenced artifact. I've inspired myself in the `oras` CLI, [which has a `--include-subject` flag](https://oras.land/docs/commands/oras_pull#examples). I've tried to inspire myself in their `pull` implementation as well however, this would require a quite substantial rewrite of the `oras-py` pull logic - `oras` CLI uses a graph approach to collect blobs to be pulled, while `oras-py` simply iterates through the layers list. Therefore I've resorted to a recursive `pull` call based on the subject digest. This may raise some potential issues in the future since, well... it's recursion and it's synchronous. 🤷

This PR depends on https://github.com/oras-project/oras-py/pull/131

### DIFF
--- a/docs/getting_started/user-guide.md
+++ b/docs/getting_started/user-guide.md
@@ -489,6 +489,28 @@ def push(uri, root):
 
 </details>
 
+<details>
+
+<summary>Example of basic artifact attachment</summary>
+
+We are assuming an `derived-artifact.txt` in the present working directory and that there's already a `localhost:5000/dinosaur/artifact:v1` artifact present in the registry. Here is an example of how to [attach](https://oras.land/docs/concepts/reftypes/) a derived artifact to the existing artifact.
+
+```python
+import oras.client
+import oras.provider
+
+client = oras.client.OrasClient(insecure=True)
+
+manifest = client.remote.get_manifest(f"localhost:5000/dinosaur/artifact:v1")
+subject = oras.provider.Subject.from_manifest(manifest)
+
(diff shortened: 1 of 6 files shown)


docs/getting_started/user-guide.md
```diff
@@ -489,6 +489,28 @@ def push(uri, root):
 
 </details>
 
+<details>
+
+<summary>Example of basic artifact attachment</summary>
```
vsoch, line 494: This looks to be redundant with the other PR - let's do that one first, merge and rebase here, and then we can do the review.

tumido: Yeah, I know... that's why I left this PR as a draft for now. 🙂  It builds on top of the other PR. I wanted to share the direction with you but wanted to keep PRs atomic at the same time. 🙂 Let's finish the first PR first and I'll rebase on top of it.

vsoch: Sounds good! Thank you for your work on both, and looking forward to coming back to this one.

PR #137, opened 2024-06-01
add arg to enable blob chunking and allow custom chunk sizes

brianwcook:
Blob chunking is necessary for uploading very large artifacts.  Quay.io, for example, can store very large artifacts but will not accept a single blob put over 20GB. Chunking support existed in the code but there was no way to activate it. The code comments indicated that it would be automatically enabled for blob sizes over 1024 bytes but no code to do that actually existed.

This code adds args to all the relevant functions starting with OrasClient.Push to allow explicit enablement of blob chunking. It also increases the fault chunk size significantly. The 1024 byte chunk size which is currently present only ensures that the data transfer will be extremely slow.  For me it tested at around 10 KB/S).  The new value results in transfer speeds of more like 10 MB/S.  Finally the blob size is also user specifiable via arguments.  Larger chunk sizes can result in faster uploads in some cases.

To test:

`import oras.client

client = oras.client.OrasClient()
client.push(files=["random.txt"], target="quay.io/bcook/tuneablechunks-b:1", do_chunked=True, chunk_size=10000000)
`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ oras.egg-info/
 env
 __pycache__
 .python-version
+/venv
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -219,13 +219,15 @@ class Registry:
         container: container_type,
         layer: dict,
         do_chunked: bool = False,
+        chunk_size=None,
         refresh_headers: bool = True,
     ) -> requests.Response:
         """
         Prepare and upload a blob.
 
-        Sizes > 1024 are uploaded via a chunked approach (post, patch+, put)
-        and <= 1024 is a single post then put.
+        Large artifacts can be uploaded via a chunked approach (post, patch+, put)
+        to registries that support it. Larger chunks generally give better throughput.
+        Set do_chunked=True for chunked upload.
 
         :param blob: path to blob to upload
         :type blob: str
@@ -233,6 +235,7 @@ class Registry:
         :type container: oras.container.Container or str
         :param layer: dict from oras.oci.NewLayer
         :type layer: dict
+        :param do_chunked: if true use chunked upload. This allows upload of larger oci artifacts.
         :param refresh_headers: if true, headers are refreshed
         :type refresh_headers: bool
         """
@@ -250,8 +253,13 @@ class Registry:
                 blob, container, layer, refresh_headers=refresh_headers
             )
         else:
+            print("chunked upload ON.")
             response = self.chunked_upload(
-                blob, container, layer, refresh_headers=refresh_headers
+                blob,
+                container,
+                layer,
+                refresh_headers=refresh_headers,
+                chunk_size=chunk_size,
             )
 
         # If we have an empty layer digest and the registry didn't accept, just return dummy successful response
@@ -556,6 +564,7 @@ class Registry:
         container: oras.container.Container,
         layer: dict,
         refresh_headers: bool = True,
+        chunk_size=None,
     ) -> requests.Response:
         """
         Upload via a chunked upload.
@@ -568,7 +577,14 @@ class Registry:
         :type layer: dict
         :param refresh_headers: if true, headers are refreshed
         :type refresh_headers: bool
+        :param chunk_size: chunk size in bytes
+        :type chunk_size: int
         """
+
+        default_chunk_size = 16777216
+        if chunk_size is None:
+            chunk_size = default_chunk_size
+
         # Start an upload session
         headers = {"Content-Type": "application/octet-stream", "Content-Length": "0"}
         if not refresh_headers:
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ oras.egg-info/
 env
 __pycache__
 .python-version
+/venv
```
vsoch, line 9: I don't think you need the leading slash for venv in the local directory.

oras/provider.py
```diff
@@ -250,8 +253,13 @@ def upload_blob(
                 blob, container, layer, refresh_headers=refresh_headers
             )
         else:
+            print("chunked upload ON.")
```
vsoch, line 256: If you want to add logging, please use the oras.logger.

oras/provider.py
```diff
@@ -219,20 +219,23 @@ def upload_blob(
         container: container_type,
         layer: dict,
         do_chunked: bool = False,
+        chunk_size=None,
```
vsoch, line 222: This should default to the default chunk size instead of None. If it's a shared value, put in defaults.py and reference from both / several function inputs.

oras/provider.py
```diff
@@ -568,7 +577,14 @@ def chunked_upload(
         :type layer: dict
         :param refresh_headers: if true, headers are refreshed
         :type refresh_headers: bool
+        :param chunk_size: chunk size in bytes
+        :type chunk_size: int
         """
+
+        default_chunk_size = 16777216
```
vsoch, line 584: Ditto here - this should be the default size, from defaults.py, in the function.

oras/provider.py
```diff
@@ -250,8 +253,13 @@ def upload_blob(
                 blob, container, layer, refresh_headers=refresh_headers
```
vsoch, line 253: Please see the comment above:

```
 # Chunked for large, otherwise POST and PUT
# This is currently disabled unless the user asks for it, as
# it doesn't seem to work for all registries
```
Chunked was not working well for different registries so I am hesitant to add it back.

oras/provider.py
```diff
@@ -585,7 +601,9 @@ def chunked_upload(
         # Read the blob in chunks, for each do a patch
         start = 0
         with open(blob, "rb") as fd:
-            for chunk in oras.utils.read_in_chunks(fd):
+            for chunk in oras.utils.read_in_chunks(fd, chunk_size=chunk_size):
+                print("uploading chunk starting at " + str(start))
```
vsoch, line 605: Please use the oras.logger here.

oras/provider.py
```diff
@@ -219,20 +219,23 @@ def upload_blob(
...
+        to registries that support it. Larger chunks generally give better throughput.
+        Set do_chunked=True for chunked upload.
 
         :param blob: path to blob to upload
         :type blob: str
         :param container:  parsed container URI
         :type container: oras.container.Container or str
         :param layer: dict from oras.oci.NewLayer
         :type layer: dict
+        :param do_chunked: if true use chunked upload. This allows upload of larger oci artifacts.
```
vsoch, line 238: If the caller wants to do a chunked upload, they can use the function directly.

oras/provider.py
```diff
@@ -689,6 +707,10 @@ def push(self, *args, **kwargs) -> requests.Response:
         :type target: str
         :param refresh_headers: if true or None, headers are refreshed
         :type refresh_headers: bool
+        :param do_chunked: if true do chunked blob upload
```
vsoch, line 710: This would better be exposed through a separate `push_chunked` or similar function, if it's a  good idea to add it back.

vsoch (commented):
Thanks @dev-zero - the chunked upload is not currently exposed because it was flaky for different registries. If you want to move forward, you'll want to rebase against master (the extra provider.py/client.py files were refactored) and then I added some comments to address. Thanks!

brianwcook:
I understand about the flaky registry behavior. We have emerging use cases that use file sizes which are too large to upload with one PUT. I am working with Quay.io engineering to ensure that chunked-upload works correctly (it currently does not) but one impediment is lack of support for chunked upload in clients to reproduce errors.  Do you have a convention for marking something experimental or unstable?  I am not suggesting that chunked upload be automatically enabled for any push; the user will need to purposefully select that mode.

oras/provider.py
brianwcook: I don't understand this comment. It would require duplicating ~ 150 lines of code which  would be otherwise unchanged. Is that really what you want?

brianwcook: Alternatively, if you like the separate function approach better than using an argument, I can rename existing 'push' to '_push' and create new, small 'push' and 'push_chunked'. Let me know how I should proceed.

vsoch: The initial design I had was fairly simple: https://github.com/oras-project/oras-py/blob/07ae3a1b46b245cb59dd390b50f3a900b07cd861/oras/provider.py#L147-L170 and based on size. I think we'd want to try a similar approach, with a boolean that defaults to False. For my comment above, I do think we want a clean separation of the two functionalities (akin to how it was done in the link above).

isinyaaa:
@vsoch @brianwcook this can be closed :)

PR #139, opened 2024-06-05
docs: use valid media_type format in examples

tarilabs:
Hi 👋 
proposing with this PR to make documentation media_type examples adhere to the MIME type/Media Type format specification.

Some OCI registry (such as Zot) seems to me they enable by default the validation of media-type according to the standard.
The current examples in the doc don't respect the standard, making the examples fail at runtime.
e.g.
```
{"level":"error","error":"[layers.0.mediaType: Does not match pattern '^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}/[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}$']","goroutine":242,"caller":"zotregistry.dev/zot/pkg/storage/common/common.go:87","time":"2024-06-05T14:26:36.160643+02:00","message":"failed to validate OCIv1 image manifest schema"}
```

The proposed modification in this PR make use of media_type examples which adhere to the MIME type/Media Type format spec, thus avoiding the validation error from the server/OCI registry side.

### DIFF
--- a/docs/getting_started/user-guide.md
+++ b/docs/getting_started/user-guide.md
@@ -132,7 +132,7 @@ import oras.defaults
 
 layers = []
 for blob in blobs:
-    layer = oras.oci.NewLayer(blob, is_dir=False, media_type="org.dinosaur.tools.blobish")
+    layer = oras.oci.NewLayer(blob, is_dir=False, media_type="application/x-org.dinosaur.tools.blobish")
 
     # This is important so oras clients can derive the relative name you want to download to
     # Using basename assumes a flat directory of files - it doesn't have to be.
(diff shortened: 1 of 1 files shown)


vsoch:
This looks good @tarilabs - not sure why the DCO isn't turning green but that's the one thing we are waiting on.

tarilabs:
> This looks good @tarilabs - not sure why the DCO isn't turning green but that's the one thing we are waiting on.

thanks @vsoch I've bumped the commit message maybe it was missing a double empty line, not sure, but seems to be passing now 👍 🚀

vsoch (approved):

PR #146, opened 2024-06-25
core: make push response the one pertaining to Manifest (not config

tarilabs:
Hi 👋 

I think it would be more helpful for Oras Registry push to return the response pertaining to the (global) Manifest of the push-action --rather then the response only of the config layer, as it's today.

1. step 1/2 with 9702a840ffef4d6712926c8c1cfff26373bddbcc shows test case (would fail on main with "Manifest not found" by means of the SHA)
2. step 2/2 with f6db11a40979252bb4ec26b36c5c0d2ae84d2e29 make the necessary implementation change (trivial change actually)

What do you think?
Hope this is helpful 🙏

### DIFF
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -805,7 +805,10 @@ class Registry:
 
         # Final upload of the manifest
         manifest["config"] = conf
-        self._check_200_response(self.upload_manifest(manifest, container))
+        response = self.upload_manifest(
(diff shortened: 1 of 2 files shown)


vsoch (approved):
This seems reasonable, and like a better overall design - thanks!

Issue #147, opened 2024-06-27
Testing registries for release 0.2.0


vsoch:
The current main branch has the TBA 0.2.0 release. I was waiting on getting a bit more testing of the new auth backend across registries of interest. Here is the list:

 - [ ] ghcr.io
 - [ ] GitLab packages
 - [ ] Docker Hub (distribution)
 - [ ] Quay.io
 - [ ] AWS
 - [ ] Azure
 - [ ] Google (gcr.io)
 - [ ] Zot
 - [ ] Quay

tarilabs:
I would like to contribute some testings, and some questions, please?

## Distribution, Basic-on-http

I believe I was able to test Basic-on-http using "distribution" (`ghcr.io/oras-project/registry:latest`) as demonstrated by this example [notebook](https://github.com/tarilabs/demo20240629-oras147-testregistries/blob/main/demo-localhost-distribution-basic.ipynb), which uses [this example configuration](https://github.com/tarilabs/demo20240629-oras147-testregistries/blob/main/config/distribution/run_registry.sh) with local "distribution" container.



## Zot, Basic-on-http

I believe I was able to test Basic-on-http using "Zot" (`ghcr.io/project-zot/zot-linux-arm64:latest`) as demonstrated by this [example notebook](https://github.com/tarilabs/demo20240629-oras147-testregistries/blob/main/demo-localhost-zot-basic.ipynb), which uses [this example configuration](https://github.com/tarilabs/demo20240629-oras147-testregistries/blob/main/config/zot/run_registry.sh) with local container.



## Distribution/Zot, using Token
I'm not sure if/how possible to test Token-on-http with either Distribution or Zot.



## Using Token
I'm having problem with Token-on-http/-https.
I've tried:
- localhost-quay using `http`: https://github.com/tarilabs/demo20240629-oras147-testregistries/blob/main/demo-localhost-quay-basic.ipynb
- quay.io using `https`: https://github.com/tarilabs/demo20240629-oras147-testregistries/blob/main/demo-quayio-token.ipynb

The error looks similar and getting Unauthorized when allegedly the token get used to upload blob.

To check the server should answer as expected, I've used with the `oras` CLI (from Homebrew) and I'm able to successfully login, then push, to both localhost-quay using `http` (using `--insecure` on the command line) and quay.io using `https` ([proof](https://quay.io/repository/mmortari/demo20240629-oras147?tab=tags)).

I don't recall this was a problem pre `<0.2.0`.

[...]

PR #148, opened 2024-07-06
core: improve anon/auth token logic

tarilabs:
while testing for #147 with ghcr.io and quay.io (so I could test the Token auth backend), I noticed the following and I believe these modification improve on the token backend logic.

I might have missed on something however 😅 , so kindly be patience and let me know if I'm not considering any gap!

Example with ghcr.io: https://github.com/users/tarilabs/packages/container/package/demo20240629-oras147
Example with quay.io: https://quay.io/repository/mmortari/demo20240629-oras147?tab=tags

To make the review hopefully easier, I've left the changes in separate commits, with richer description.

wdyt?

(note: c243eb2 introduced later as in #147 notebooks I was only testing auth backends for push and pull)

### DIFF
--- a/oras/auth/token.py
+++ b/oras/auth/token.py
@@ -44,7 +44,7 @@ class TokenAuth(AuthBackend):
             self.set_header("Authorization", "Basic %s" % self._basic_auth)
 
     def authenticate_request(
-        self, original: requests.Response, headers: dict, refresh=False, skipAnonToken=False
+        self, original: requests.Response, headers: dict, refresh=False
     ):
         """
         Authenticate Request
@@ -72,20 +72,17 @@ class TokenAuth(AuthBackend):
         h = auth_utils.parse_auth_header(authHeaderRaw)
 
         # First try to request an anonymous token
-        if not skipAnonToken:
-            logger.debug("No Authorization, requesting anonymous token")
-            anon_token = self.request_anonymous_token(h)
-            if anon_token:
-                logger.debug("Successfully obtained anonymous token!")
-                self.token = anon_token
-                headers["Authorization"] = "Bearer %s" % self.token
-                return headers, True
+        logger.debug("No Authorization, requesting anonymous token")
+        anon_token = self.request_anonymous_token(h)
+        if anon_token:
+            logger.debug("Successfully obtained anonymous token!")
+            self.token = anon_token
+            headers["Authorization"] = "Bearer %s" % self.token
+            return headers, True
 
         # Next try for logged in token
-        logger.debug("requesting auth token")
         token = self.request_token(h)
         if token:
-            logger.debug("Successfully obtained auth token!")
             self.token = token
             headers["Authorization"] = "Bearer %s" % self.token
             return headers, True
--- a/oras/auth/token.py
+++ b/oras/auth/token.py
@@ -44,7 +44,11 @@ class TokenAuth(AuthBackend):
             self.set_header("Authorization", "Basic %s" % self._basic_auth)
 
     def authenticate_request(
-        self, original: requests.Response, headers: dict, refresh=False, skipAnonToken=False
+        self,
+        original: requests.Response,
+        headers: dict,
+        refresh=False,
+        skipAnonToken=False,
     ):
         """
         Authenticate Request
@@ -102,7 +106,7 @@ class TokenAuth(AuthBackend):
         """
         params = {}
         headers = {}
-        if self._basic_auth: # we exchange the basic auth for the token
+        if self._basic_auth:  # we exchange the basic auth for the token
             headers["Authorization"] = "Basic %s" % self._basic_auth
 
         # Prepare request to retry
--- a/oras/auth/token.py
+++ b/oras/auth/token.py
@@ -73,7 +73,6 @@ class TokenAuth(AuthBackend):
 
         # if no basic auth, try by request an anonymous token
         if not hasattr(self, "_basic_auth"):
-            logger.debug("No Basic Auth found, requesting anonymous token")
             anon_token = self.request_anonymous_token(h)
             if anon_token:
                 logger.debug("Successfully obtained anonymous token!")
@@ -81,8 +80,7 @@ class TokenAuth(AuthBackend):
                 headers["Authorization"] = "Bearer %s" % self.token
                 return headers, True
 
-        # try using auth token
-        logger.debug("requesting Auth Token")
+        # basic auth is available, try using auth token
         token = self.request_token(h)
         if token:
             self.token = token
@@ -97,7 +95,7 @@ class TokenAuth(AuthBackend):
 
     def request_token(self, h: auth_utils.authHeader) -> bool:
         """
-        Request an authenticated token and save for later.s
+        Request an authenticated token and save for later.
         """
         params = {}
         headers = {}
@@ -126,6 +124,7 @@ class TokenAuth(AuthBackend):
         # Set Basic Auth to receive token
         headers["Authorization"] = "Basic %s" % self._basic_auth
 
+        logger.debug(f"Requesting auth token for: {h}")
         authResponse = self.session.get(h.realm, headers=headers, params=params)  # type: ignore
 
         if authResponse.status_code != 200:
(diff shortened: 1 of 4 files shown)


oras/auth/token.py
```diff
@@ -44,7 +44,11 @@ def reset_basic_auth(self):
             self.set_header("Authorization", "Basic %s" % self._basic_auth)
 
     def authenticate_request(
-        self, original: requests.Response, headers: dict, refresh=False
+        self,
+        original: requests.Response,
+        headers: dict,
+        refresh=False,
+        skipAnonToken=False,
```
vsoch, line 51: Please use `skip_anon_token` to match the rest of the code base style, and since CamelCase is typically for Go/Java.

oras/auth/token.py
```diff
@@ -73,17 +77,20 @@ def authenticate_request(
...
 
         # First try to request an anonymous token
-        logger.debug("No Authorization, requesting anonymous token")
-        anon_token = self.request_anonymous_token(h)
-        if anon_token:
-            logger.debug("Successfully obtained anonymous token!")
-            self.token = anon_token
-            headers["Authorization"] = "Bearer %s" % self.token
-            return headers, True
+        if not skipAnonToken:
```
vsoch, line 80: Isn't there a more programmatic way to figure this out than provide a boolean? If they just don't provide any credentials can't we assume using anon?

oras/auth/token.py
```diff
@@ -100,6 +107,8 @@ def request_token(self, h: auth_utils.authHeader) -> bool:
         """
         params = {}
         headers = {}
+        if self._basic_auth:  # we exchange the basic auth for the token
```
vsoch, line 110: Wouldn't this be a separate part of basic auth?

vsoch (commented):
The tests look good! Some quick questions.

oras/auth/token.py
tarilabs: oops sorry 😅  will do 👍

oras/auth/token.py
tarilabs: My understanding of this section of the token backend is:
- We want to exchange a basic auth for an auth token
- to do so, we need to supply our basic auth to the realm
- the realm, seeing our basic auth credential, will give us an auth bearer token (which we could use in the subsequent requests)

so haven't fully understood how you would rewrite this part? 🤔 can you expand it a little bit more for me please?

oras/auth/token.py
tarilabs: that it's a nice idea, could experiment with it, based on feedback from: https://github.com/oras-project/oras-py/pull/148#discussion_r1667623649

oras/auth/token.py
vsoch: Gotcha - I think I am wondering what the failure case here is. If we didn't get an anon or authenticated token, then the auth backend remains basic. Should we allow that or fail somewhere with a targeted message?

tarilabs: mmm... the reason for this changed line is different: this is method for requiring auth-token, but was not providing any auth, it was requesting a token which --for all intent and purposes-- was analogous to the anon-token requested by the other method.

I could verify in my case the experienced behaviour was:
- from provider, 1st request (no auth at all) failed
- from provider, delegate to auth backend to perform authentication
- the token auth backend, requested anon token
- the token auth backend, receives from server "I won't give you any anon token"
- the token auth backend, moves on to request an auth-token
- _before this code change_ the token auth backed did not provide any basic auth to the Realm, so in effect it was analogous to the anon token just failed --because servers such as ghcr.io do not offer anon tokens
- _after this code change_ the token auth backend provides the basic auth to the Realm, and the Realm provides back the auth-token successfully

This is my understanding on how the flow is meant to happen, hence why this line change.
If I missed something, kindly let me know!

oras/auth/token.py
tarilabs: done with 102381c

oras/auth/token.py
tarilabs: > If they just don't provide any credentials can't we assume using anon?

I would like to double-check I'm understanding your suggestion correctly, please.

In the current main branch code (unrelated to this PR), the anon token is always tried first, and as far as I can tell, there is no way when the second call to `token.py authenticate_request()` happens to  check whether the anon token was tried, it actually always re-tries the anon token:

https://github.com/oras-project/oras-py/blob/caf8db5b279382335fbb1f6d7402ed9b73618d37/oras/auth/token.py#L46-L85

See line 77.
It appears to me, line 77 is always executed for asking a `anon_token = self.request_anonymous_token(h)  `
and there is no way to jump down to line 85, to request the auth token.

This is what appears to me is happening today on main branch.

With this PR, I've introduced a deliberate flag whether anon_token should be skipped (so avoid line 77, and jump to 85) and the 3rd attempt from `provider.py do_request()`.

What your suggestion now makes me wonder now, is whether to actually simplify the logic differently here as:
- anon token request is only tried if there is NO auth entry
- in case an auth entry is available, never try anon token, and just perform auth token

Can you kindly feedback @vsoch about this, please?
Thank you, and hope this helps.

vsoch: > it actually always re-tries the anon token

For request token, the basic auth is needed right? So perhaps that's the logic that should be added before showing the user:

```python
logger.debug("No Authorization, requesting anonymous token") 
```
And if the basic auth is there, skip over asking for an anon token. If there are cases where we should try the basic auth first and fall back to anonymous when needed, that could be done too.

tarilabs: implemented with https://github.com/oras-project/oras-py/pull/148/commits/bb4c2ec85f8614dc9a21e3c9ea7d9feb388ff91c

tarilabs:
up to bb4c2ec (next 9c3f003 is just linting) been using [this external project](https://github.com/tarilabs/demo20240629-oras147-testregistries/commit/5921e38a7ea0de151a460dbcaf39f133a49f1212) to test it works as expected exchanging for the request token, examples:

- ghcr.io:
image

- quay.io:
image
 

I think this incorporates needed changes to avoid anon tokens when a request token should be used instead, required to my knowledge and exemplified above.

In the longer term, I'm thinking to contribute some automated tests, as suggested https://github.com/oras-project/oras-py/issues/147#issuecomment-2335099677

oras/auth/token.py
```diff
@@ -71,16 +71,18 @@ def authenticate_request(
...
+            logger.debug("No Basic Auth found, requesting anonymous token")
+            anon_token = self.request_anonymous_token(h)
+            if anon_token:
+                logger.debug("Successfully obtained anonymous token!")
+                self.token = anon_token
+                headers["Authorization"] = "Bearer %s" % self.token
+                return headers, True
+
+        # try using auth token
+        logger.debug("requesting Auth Token")
```
vsoch, line 85: Could you put this debug message in the function that does that, and put a bit more metadata to make the comment more useful?

tarilabs: done with 7d891e96e24f16f690be4c3953a53cbd819dfe6f

tarilabs:
kindly let me know if these overall changes looks good @vsoch , thank you so much for your patience on this one 🙏 hope these changes helps

oras/auth/token.py
```diff
@@ -152,7 +151,7 @@ def request_anonymous_token(self, h: auth_utils.authHeader) -> bool:
         if h.scope:
             params["scope"] = h.scope
 
-        logger.debug(f"Final params are {params}")
+        logger.debug(f"Requesting anon token with params: {params}")
```
vsoch, line 154: Just a sanity check - could this have any parameters we wouldn't want to be printed? I would probably leave this line out entirely.

tarilabs: only `service` and `scope`:

https://github.com/tarilabs/oras-py/blob/19f452836513c572c248027352a5d95eaefd4337/oras/auth/token.py#L148-L152

which to my understanding are not sensitive info and done analogously for the auth token.

Just to be clear, I've just changed the hardcoded text to make it significant it's about an anon token when observing the logs, and the `params` was there before :)

I thought aligning log statements in:

- https://github.com/tarilabs/oras-py/blob/7d891e96e24f16f690be4c3953a53cbd819dfe6f/oras/auth/token.py#L127-L128
- https://github.com/tarilabs/oras-py/blob/7d891e96e24f16f690be4c3953a53cbd819dfe6f/oras/auth/token.py#L154-L155

to your comment of:

> ... put this debug message in the function that does that, and put a bit more metadata to make the comment more useful ...

([source](https://github.com/oras-project/oras-py/pull/148#discussion_r1771573650))

but let me know if you prefer me to change those log debug statement to something else!

vsoch: > only service and scope:

I just wanted to double check this - I think this should be ok to print.

vsoch (approved):

tarilabs:
thank you so much @vsoch 🙏 🚀 🪐

PR #150, opened 2024-08-19
Enable chunked uploads

isinyaaa:
TL;DR: Rebases #137. Also fixes out-of-date `_state_` parameter on `session_url`, which caused a 404 when resuming/completing uploads.

We wanted to use oras for larger uploads (say, ML model files) at containers/omlmd, but I wasn't able to make them work with standard uploads. I noticed #137, rebased it, addressed some of your comments (not sure how to address all of them, though). But I still couldn't make it work with https://hub.docker.com/_/registry. So I started debbuging to find out there's a `_state_` parameter being passed around, and I assume it must be updated on the https://distribution.github.io/distribution/spec/api/#completed-upload PUT request to reflect the last state reported by the server. I tested this with 20GB-ish files.

I wonder if this could help with bringing back chunked file support by default, although I'm not really sure in which aspect that kind of support is "flaky" or (as I experienced) poorly documented.

### DIFF
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -598,6 +598,7 @@ class Registry:
         """
         # Start an upload session
         headers = {"Content-Type": "application/octet-stream", "Content-Length": "0"}
+        headers.update(self.headers)
 
         upload_url = f"{self.prefix}://{container.upload_blob_url()}"
         r = self.do_request(upload_url, "POST", headers=headers)
@@ -611,11 +612,6 @@ class Registry:
         start = 0
         with open(blob, "rb") as fd:
             for chunk in oras.utils.read_in_chunks(fd, chunk_size=chunk_size):
-                print("uploading chunk starting at " + str(start))
-
-                if not chunk:
-                    break
-
                 end = start + len(chunk) - 1
                 content_range = "%s-%s" % (start, end)
                 headers = {
@@ -623,13 +619,19 @@ class Registry:
                     "Content-Length": str(len(chunk)),
                     "Content-Type": "application/octet-stream",
                 }
+                headers.update(self.headers)
 
                 # Important to update with auth token if acquired
                 # TODO call to auth here
                 start = end + 1
                 self._check_200_response(
-                    self.do_request(session_url, "PATCH", data=chunk, headers=headers)
+                    r := self.do_request(
+                        session_url, "PATCH", data=chunk, headers=headers
+                    )
                 )
+                session_url = self._get_location(r, container)
+                if not session_url:
+                    raise ValueError(f"Issue retrieving session url: {r.json()}")
 
         # Finally, issue a PUT request to close blob
         session_url = oras.utils.append_url_params(
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -251,12 +251,14 @@ class Registry:
         container: container_type,
         layer: dict,
         do_chunked: bool = False,
+        chunk_size: int = oras.defaults.default_chunksize,
     ) -> requests.Response:
         """
         Prepare and upload a blob.
 
-        Sizes > 1024 are uploaded via a chunked approach (post, patch+, put)
-        and <= 1024 is a single post then put.
+        Large artifacts can be uploaded via a chunked approach (post, patch+, put)
+        to registries that support it. Larger chunks generally give better throughput.
+        Set do_chunked=True for chunked upload.
 
         :param blob: path to blob to upload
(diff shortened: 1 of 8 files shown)


oras/provider.py
```diff
@@ -596,24 +611,27 @@ def chunked_upload(
         # Read the blob in chunks, for each do a patch
         start = 0
         with open(blob, "rb") as fd:
-            for chunk in oras.utils.read_in_chunks(fd):
-                if not chunk:
```
vsoch, line 600: Why did you remove this check?

vsoch (commented):
This looks good for testing - could you please add / re-enable a test for chunked?

oras/provider.py
isinyaaa: Even if your iterator doesn't raise, it stops yielding after a certain point, and there's an internal check with the same condition too. Why should it be necessary to check here again?

vsoch: You're right - I forgot that I added that - it must have not been there at some point. Reference:

https://github.com/oras-project/oras-py/blob/caf8db5b279382335fbb1f6d7402ed9b73618d37/oras/utils/fileio.py#L260-L273

vsoch:
@isinyaaa please see my previous review comment - we need explicit tests for the chunked upload.

isinyaaa:
> @isinyaaa please see my previous review comment - we need explicit tests for the chunked upload.

@vsoch sorry I missed that. Updated now, wdyt?

vsoch:
Nice! Let's run these tests now.

isinyaaa:
@vsoch can you reapprove? I ran the linter locally to make sure it's working, sorry for the trouble

isinyaaa:
@vsoch not sure what went wrong with those tests... maybe some issue with the generated file size? from the raw logs I can't spot any problems, neither locally.

vsoch:
Here is what I see:

```

/bin/bash scripts/test.sh
ORAS_PORT: 5000
ORAS_HOST: localhost
ORAS_REGISTRY: localhost:5000
ORAS_AUTH: 
============================= test session starts ==============================
platform linux -- Python 3.11.9, pytest-8.3.2, pluggy-1.5.0
rootdir: /home/runner/work/oras-py/oras-py
configfile: pyproject.toml
collected 22 items
oras/tests/test_oci.py .
oras/tests/test_oras.py .sSuccessfully pushed localhost:5000/dinosaur/artifact:v1
Successfully pushed localhost:5000/dinosaur/artifact:v1
.Successfully pushed localhost:5000/dinosaur/artifact:v1
.Successfully pushed localhost:5000/dinosaur/artifact:v1
..Successfully pushed localhost:5000/dinosaur/directory:v1
.s
oras/tests/test_provider.py Successfully pushed localhost:5000/dinosaur/artifact:v1
Successfully pushed localhost:5000/dinosaur/artifact:v1
0+0 records in
0+0 records out
0 bytes copied, 3.8763e-05 s, 0.0 kB/s
```
I would try to reproduce locally.

isinyaaa:
@vsoch as expected, while testing on my fork I found the problem lies when working with those very large files on GHA, reduced the test file size to be a couple times the default chunk size, wdyt?

vsoch:
Should the chunk size perhaps be smaller then?

isinyaaa:
I don't really think that's a problem, up to you. The issue was in creating a 15GB test file in github actions. I think the worker didn't have this much space to spare or something. I reduced the test file to be 4x the default chunk size, that's all.

vsoch:
Ah ok. Please keep the tests in GitHub the same as what you are doing (and what is working) locally and let's try making more space on the builder - these first three lines before the tests to cleanup and add space should be sufficient:

https://github.com/converged-computing/fluxnetes/blob/0d577aa3155e68aff457f390f8c926e9f57a13d3/.github/workflows/e2e-test.yaml#L129-L133

But you can add more as needed.

isinyaaa:
@vsoch updated! thanks for the pointer, I had no idea the default images could be so huge! Though the first three lines didn't suffice as we actually need 15GB for the file + 15GB for the upload.

isinyaaa:
Hey, @vsoch, can we merge this yet?

vsoch:
Yes - we are close! Can you please bump the version in oras/version.py and make a note about the change in CHANGELOG.md? That should be the final bit we need.

isinyaaa:
updated, wdyt?

vsoch (approved):

PR #152, opened 2024-09-04
Make pyright happy

poksiala:
I tried to use this library and to be honest the user experience was not the best. For me it was hard to understand with all the missing and invalid type annotations. My editor was also complaining a lot, especially about the decorated methods that did not pass trough information about the methods.

So as an exercise to better understand the inner workings of the library I started to fix those type annotations. I also found some dead code and missing variables while doing that. Here are the results if you are interested in including them in the trunk.

This should not change the API at all though I have had no time to test it in actual use yet.

### DIFF
--- a/oras/auth/base.py
+++ b/oras/auth/base.py
@@ -2,37 +2,49 @@ __author__ = "Vanessa Sochat"
 __copyright__ = "Copyright The ORAS Authors."
 __license__ = "Apache-2.0"
 
-
-from typing import Optional
+import abc
+from typing import Dict, Optional, Tuple
 
 import oras.auth.utils as auth_utils
 import oras.container
-import oras.decorator as decorator
 from oras.logger import logger
-from oras.types import container_type
+import oras.utils
+
+import requests
 
 
-class AuthBackend:
+class AuthBackend(abc.ABC):
     """
     Generic (and default) auth backend.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, session: requests.Session):
         self._auths: dict = {}
+        self.session = session
+        self.headers: Dict[str, str] = {}
+
+    @abc.abstractmethod
+    def authenticate_request(
+        self,
+        original: requests.Response,
+        headers: Optional[dict[str, str]] = None,
+        refresh=False,
+    ) -> Tuple[dict[str, str], bool]:
+        pass
+
+    def set_header(self, name: str, value: str):
+        """
+        Courtesy function to set a header
+
+        :param name: header name to set
+        :type name: str
+        :param value: header value to set
+        :type value: str
+        """
+        self.headers.update({name: value})
 
     def get_auth_header(self):
         raise NotImplementedError
-
-    def get_container(self, name: container_type) -> oras.container.Container:
-        """
-        Courtesy function to get a container from a URI.
-
-        :param name: unique resource identifier to parse
-        :type name: oras.container.Container or str
-        """
-        if isinstance(name, oras.container.Container):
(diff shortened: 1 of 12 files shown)


oras/auth/base.py
```diff
@@ -2,37 +2,49 @@
...
+        original: requests.Response,
+        headers: Optional[dict[str, str]] = None,
+        refresh=False,
+    ) -> Tuple[dict[str, str], bool]:
+        pass
 
-    def get_auth_header(self):
-        raise NotImplementedError
-
-    def get_container(self, name: container_type) -> oras.container.Container:
```
poksiala, line 26: I removed this method as it relies on `self.hostname` which is not available

oras/auth/base.py
```diff
@@ -81,8 +93,9 @@ def _load_auth(self, hostname: str) -> bool:
             return True
         return False
 
-    @decorator.ensure_container
-    def load_configs(self, container: container_type, configs: Optional[list] = None):
+    def load_configs(
+        self, container: oras.container.Container, configs: Optional[list] = None
```
poksiala, line 97: I changed the container type as it relied on `self.get_container()` method above trough the decorator. Which as said would not have worked anyways

oras/auth/base.py
```diff
@@ -120,37 +133,3 @@ def set_basic_auth(self, username: str, password: str):
         :type password: str
         """
         self._basic_auth = auth_utils.get_basic_auth(username, password)
-
-    def request_anonymous_token(self, h: auth_utils.authHeader, headers: dict) -> bool:
```
poksiala, line 124: To my understanding this is dead code and completely incompatible with the overriding method in token auth, which is actually being used.

oras/auth/base.py
```diff
@@ -2,37 +2,49 @@
...
+        headers: Optional[dict[str, str]] = None,
+        refresh=False,
+    ) -> Tuple[dict[str, str], bool]:
+        pass
 
-    def get_auth_header(self):
-        raise NotImplementedError
-
-    def get_container(self, name: container_type) -> oras.container.Container:
+    def set_header(self, name: str, value: str):
```
poksiala, line 35: This method was referenced in multiple places but did not actually exist. So I added it by copying it from Provider.

oras/auth/token.py
```diff
@@ -113,16 +120,18 @@ def request_token(self, h: auth_utils.authHeader) -> bool:
                 }
             )
 
+        assert h.realm is not None, "realm must be defined"
+
         # Ensure the realm starts with http
-        if not h.realm.startswith("http"):  # type: ignore
-            h.realm = f"{self.prefix}://{h.realm}"
+        if not h.realm.startswith("http"):
+            h.realm = f"http://{h.realm}"  # TODO: Should this be htts
```
vsoch, line 127: https?

oras/provider.py
```diff
@@ -120,10 +122,9 @@ def delete_tags(self, name: str, tags=Union[str, list]) -> List[str]:
         :param tags: single or multiple tags name to delete
         :type N: string or list
         """
-        if isinstance(tags, str):
-            tags = [tags]
+        _tags = [tags] if isinstance(tags, str) else tags
```
vsoch, line 125: Why did you make these variables private?

oras/auth/token.py
poksiala: Good catch! fixed

vsoch:
@poksiala apologies this fell off my radar - if you want to rebase I can give another review and (I think) it should be ready for merge.

Issue #154, opened 2024-09-14
logger.exit(...)


xarses:
I recently upgraded from 0.1.17 to 0.2.2, and found that there are two cases  in the internals ([here](https://github.com/oras-project/oras-py/blob/main/oras/auth/__init__.py#L14) and [here](https://github.com/oras-project/oras-py/blob/main/oras/provider.py#L978)) where `logger.exit(...)`[is used](https://github.com/oras-project/oras-py/blob/main/oras/logger.py#L214) which causes a `sys.exit(1)` rather than raising an exception. Is there a reasoning for handling the error(s) this way instead of letting the caller handle these?

vsoch:
We should raise a ValueError instead (with the same message). Would you care to open a PR to make this change?

PR #155, opened 2024-09-14
Replace logger.exit(...) from the internals

xarses:
Replace with raise ValueError instead so that the caller has the option to handle the problem as they desire.

Fixes: #154

### DIFF
--- a/oras/auth/__init__.py
+++ b/oras/auth/__init__.py
@@ -1,7 +1,5 @@
 import requests
 
-from oras.logger import logger
-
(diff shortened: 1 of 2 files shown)


vsoch (approved):

vsoch:
@xarses you will need to sign the DCO to merge this fully - click on the little "Details" next to the Robot that says "DCO" in the checks section.

xarses:
just signed-off-by? should be good. Thanks!

PR #158, opened 2024-09-17
Add a dedicated Exemption for auth handlers

xarses:
In 4d248a8 we removed the sys.exit(...) but that now allows for auth failures to be retried by the decorators. Adding a dedicated Exemption so we don't have to retry on fatal auth-failures

### DIFF
--- a/oras/auth/__init__.py
+++ b/oras/auth/__init__.py
@@ -6,6 +6,12 @@ from .token import TokenAuth
 auth_backends = {"token": TokenAuth, "basic": BasicAuth}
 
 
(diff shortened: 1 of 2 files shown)


oras/auth/__init__.py
```diff
@@ -6,6 +6,12 @@
 auth_backends = {"token": TokenAuth, "basic": BasicAuth}
 
 
+class AuthenticationException(Exception):
+    """
+    An exception to traise with Authentication errors are fatal
```
vsoch, line 11: Would be good to fix the typo here, and just for readability add a pass.

Issue #164, opened 2024-10-17
Issue with accessing a dockerhub registery


shaunak-cisco:
I am trying to run the basic pull example provided here: https://oras-project.github.io/oras-py/getting_started/user-guide.html with the python sdk. Following is the code snippet I am trying to run. The file getting pulled is already present in the repo pushed with the `oras` cli. I have tried pulling with the cli as well as a Golang SDK [example](https://pkg.go.dev/oras.land/oras-go/v2#example-package-PullFilesFromRemoteRepository) which worked fine.

```
import oras.client
from oras.logger import setup_logger, logger
setup_logger(quiet=False, debug=True)

password = "<>"
user = "<>"
host = "dockerhub.<>.com"
client = oras.client.OrasClient()

print(client.login(username=user, password=password, hostname=host))
client.pull(target="dockerhub.<>.com/<>/<>:<version>")
```

Following is a condensed snippet of the exception:

```
{'Status': 'Login Succeeded'}
Retrying in 3 seconds - error: 'TokenAuth' object has no attribute 'prefix'
Retrying in 5 seconds - error: 'TokenAuth' object has no attribute 'prefix'
Retrying in 11 seconds - error: 'TokenAuth' object has no attribute 'prefix'
Retrying in 29 seconds - error: 'TokenAuth' object has no attribute 'prefix'
Retrying in 83 seconds - error: 'TokenAuth' object has no attribute 'prefix'

AttributeError: 'TokenAuth' object has no attribute 'prefix'
```
Looking at the above logs, the `prefix` attribute seems to get referred in `/oras/auth/token.py:L117` but I could not see where it is getting set in the codebase.

Any pointers on if my code may not be accurate here?

tarilabs:
can you kindly confirm which version of oras-py are you using?

btw fwiw and maybe it could help, it works for me with latest oras-py for push/pull for the public dockerhub: https://hub.docker.com/repository/docker/matteomortari/demo20241017-oraspy164
with the following snippet:

```py 
# ...
host = "registry-1.docker.io"
client = oras.client.OrasClient()

print(client.login(username=user, password=password, hostname=host))
client.push(target="registry-1.docker.io/matteomortari/demo20241017-oraspy164:latest", files=["artifact.txt"])
client.pull(target="registry-1.docker.io/matteomortari/demo20241017-oraspy164:latest", outdir="tmp")
```

Using user and password per configured PAT which I've also tested from oras cli (go).

---

[...]

PR #165, opened 2024-10-17
core: add missing prefix property to auth backend

tarilabs:
ref https://github.com/oras-project/oras-py/issues/164#issuecomment-2419535612

### DIFF
--- a/oras/auth/__init__.py
+++ b/oras/auth/__init__.py
@@ -14,10 +14,11 @@ class AuthenticationException(Exception):
     pass
 
 
-def get_auth_backend(name="token", session=None, **kwargs):
+def get_auth_backend(name="token", session=None, prefix="https", **kwargs):
     backend = auth_backends.get(name)
     if not backend:
         raise ValueError(f"Authentication backend {backend} is not known.")
     backend = backend(**kwargs)
     backend.session = session or requests.Session()
+    backend.prefix = prefix
     return backend
--- a/oras/auth/__init__.py
+++ b/oras/auth/__init__.py
@@ -14,11 +14,11 @@ class AuthenticationException(Exception):
     pass
 
 
-def get_auth_backend(name="token", session=None, prefix="https", **kwargs):
+def get_auth_backend(name="token", session=None, insecure=False, **kwargs):
     backend = auth_backends.get(name)
     if not backend:
         raise ValueError(f"Authentication backend {backend} is not known.")
     backend = backend(**kwargs)
     backend.session = session or requests.Session()
-    backend.prefix = prefix
+    backend.prefix = "http" if insecure else "https"
     return backend
--- a/oras/auth/base.py
+++ b/oras/auth/base.py
@@ -19,6 +19,7 @@ class AuthBackend:
 
     def __init__(self, *args, **kwargs):
         self._auths: dict = {}
+        self.prefix: str = "https"
 
     def get_auth_header(self):
         raise NotImplementedError
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -78,7 +78,7 @@ class Registry:
         self.session.cookies.set_policy(DefaultCookiePolicy(allowed_domains=[]))
 
         # Get custom backend, pass on session to share
-        self.auth = oras.auth.get_auth_backend(auth_backend, self.session)
+        self.auth = oras.auth.get_auth_backend(auth_backend, self.session, self.prefix)
 
     def __repr__(self) -> str:
         return str(self)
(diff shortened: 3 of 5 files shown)


oras/auth/__init__.py
```diff
@@ -14,10 +14,11 @@ class AuthenticationException(Exception):
     pass
 
 
-def get_auth_backend(name="token", session=None, **kwargs):
+def get_auth_backend(name="token", session=None, prefix="https", **kwargs):
```
vsoch, line 17: @tarilabs this won't always be https - specifically in testing. I think we'd want to inherit this from the parent that created it, and use the same convention there with insecure as a boolean to determine the prefix (and not hard code http or https). Let me know your thoughts.

tarilabs: this line is specifically about the default isn't it,
and is mainly never used because here:

https://github.com/oras-project/oras-py/pull/165/files#diff-ccb407ceff93417721d33c41bf1975c04c98454fd2a36518a7b8572808db03a4R81

is set explicitly. 

am I missing something? 🤔

vsoch: Yes I think you want to set it based on the insecure parameter:

```
 self.prefix: str = "http" if insecure else "https"
```
Which is in the init of the provider.

tarilabs: mmm, I've actually mimic-ed on existing code and did that in line 23 of this file.

making the line you quote:

https://github.com/oras-project/oras-py/pull/165/files#diff-fcbd6629a667f0deb29e01bccd3fa21f8790961f48092d8cee24ad00986706a9R22

another (unused) default.

I'm happy to refactor this code differently if that helps, I was just trying to go with the existing "code convention"? :)

oras/provider.py
```diff
@@ -78,7 +78,7 @@ def __init__(
         self.session.cookies.set_policy(DefaultCookiePolicy(allowed_domains=[]))
 
         # Get custom backend, pass on session to share
-        self.auth = oras.auth.get_auth_backend(auth_backend, self.session)
+        self.auth = oras.auth.get_auth_backend(auth_backend, self.session, self.prefix)
```
tarilabs, line 81: step 1.

the prefix of provider is set a few lines above:

oras.auth.get_auth_backend

and is "pass forward" as suggested to the call of `oras.auth.get_auth_backend()`

oras/auth/__init__.py
tarilabs, line 17: step 2.

unless otherwise specified, this function `get_auth_backend()` parameter `prefix` defaults to https; but for all usages in oras code, is set from provider's prefix, as we seen in step1.

oras/auth/__init__.py
```diff
@@ -14,10 +14,11 @@ class AuthenticationException(Exception):
...
     backend = auth_backends.get(name)
     if not backend:
         raise ValueError(f"Authentication backend {backend} is not known.")
     backend = backend(**kwargs)
     backend.session = session or requests.Session()
+    backend.prefix = prefix
```
tarilabs, line 23: step 3.
the determined backend (either token or basic) get `.prefix` property set to the value received in step 2

this maintains code-style for `.session` another property set on determined backend.

oras/auth/base.py
```diff
@@ -19,6 +19,7 @@ class AuthBackend:
 
     def __init__(self, *args, **kwargs):
         self._auths: dict = {}
+        self.prefix: str = "https"
```
tarilabs, line 22: step 4.

the base `AuthBackend` by default would have a property `prefix` valorized to https, but as we just seen in step 3, this is set to the actual prefix value.

tarilabs (commented):
hope the comment in https://github.com/oras-project/oras-py/pull/165/files#r1805164912 answer the last question, but if that helps, here is the run-down of how keeping the same existing code-style as present today on main, the prefix is set for the auth backed--see below steps with my reasoning out loud, and happy to be corrected!

If you want me to refactor differently, kindly let me know in which way differing from current style; thanks!

vsoch:
@tarilabs I understand your logic - the question is if a user is ever going to call `get_auth_backend` outside of the provider. In that case, they could do something like:

```python
get_auth_backend(name="token", session=None, prefix="bananas")
```
In which case my preference is to better control that, and (although there is redundancy in setting the prefix) you can better control that with:

```python
get_auth_backend(name="token", session=None, insecure=False):
    ...
    self.prefix = "http" if insecure else "https"
```

My preference is to generally not have open string arguments like that, but instead booleans that control them.

tarilabs:
thank you for taking the time to explain to me this "gap" @vsoch ; I believe I got it now; let me refactor it 👍

tarilabs:
Refactored and maintained version update and changelog, so to make a release to support conversations in #164

vsoch (approved):
This looks good! Thanks @tarilabs!

PR #168, opened 2024-11-13
Retry on 500

isinyaaa:
This is the standard policy in the upstream [oras-project/oras-go](https://github.com/oras-project/oras-go/blob/bf570bcc78b3aa1bad84401e3390cc79a03e426e/registry/remote/retry/policy.go#L56). It actually fixes an issue when uploading manifests to Quay.io, as can be observed in this HTTP traffic capture from WireShark:

image

Tested with a local instance and also with Quay.io successfully following https://github.com/isinyaaa/quay-demo.git
Remote upload https://quay.io/repository/idoamara/testrepo verified with `oras cp quay.io/idoamara/testrepo:v1 --to-oci-layout testrepo-mirror`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - retry on 500 (0.2.25)
  - align provider config_path type annotations (0.2.24)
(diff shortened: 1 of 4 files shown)


vsoch (commented):
This looks good! And makes sense. Could you please bump the version in version.py and add a corresponding note to the CHANGELOG.md about the changes?

isinyaaa:
For the record it seems that Quay manifest push is failing due to a missing property on the manifest: https://github.com/opencontainers/image-spec/blob/main/config.md#:~:text=rootfs%20object%2C%20REQUIRED
We're working on a fix for oras.

vsoch (approved):

PR #170, opened 2024-12-03
make oras-py behave the same way as oras-go for deciding whether to unpack a layer

MichaelKopfMkf:
resolves https://github.com/oras-project/oras-py/issues/119

use the same logic as oras-go and cli for determining whether to unpack a tar layer or not.

With this implementation oras-py does not anymore use the layer's media type, it uses the annotation io.deis.oras.content.unpack.

that way artifacts published by oras-cli can be consumed the same way with oras-py and vice versa.

in addition, a skip_unpack param was added to the pull command in favor of https://github.com/oras-project/oras-py/issues/119

PR #179, opened 2024-12-10
Make oras-py behave the same way as oras-go for deciding whether to unpack a layer 

### LINKED ISSUE #119: Disable tgz extraction on pull
It seems that client.pull() automatically extracts archives which is a nice feature however I don't require it so it is possible to disable?

I've also noticed that the result from the pull is an invalid path, the path is actually the path to the extacted folder however the returned path includes the tgz which is infact in another folder (all be it with a random name)

image

image

image

stefansli:
See initial implementation at #170

Fixes #119

### DIFF
--- a/oras/tests/test_oras.py
+++ b/oras/tests/test_oras.py
@@ -2,12 +2,14 @@ __author__ = "Vanessa Sochat"
 __copyright__ = "Copyright The ORAS Authors."
 __license__ = "Apache-2.0"
 
+import json
 import os
 import shutil
 
 import pytest
 
 import oras.client
+import oras.oci
 
 here = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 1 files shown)


oras/provider.py
```diff
@@ -899,9 +901,13 @@ def pull(
...
         for layer in manifest.get("layers", []):
-            filename = (layer.get("annotations") or {}).get(
-                oras.defaults.annotation_title
-            )
+            annotations: dict = layer.get("annotations", {})
+            filename = annotations.get(oras.defaults.annotation_title)
+            # A directory will need to be uncompressed and moved
+            unpack_layer = annotations.get(oras.defaults.annotation_unpack, False)
+
+            if unpack_layer and skip_unpack:
```
vsoch, line 909: How do we know the filename doesn't already have .tar.gz?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - use same annotation as oras-go for determining whether to unpack a layer or not (0.2.26)
```
vsoch, line 17: This should be nested under the group below (with same version string)

oras/provider.py
stefansli: After thinking about this for a while: 

Do we even need the `skip_unpack` parameter if we support the official annotation?
For special cases, one can always call `self.download_blob` directly.

To me `unpack_layer and skip_unpack` reads like a contradiction with behavior that is not obvious and must be documented. So I'd rather leave it out. What do you think?

PR #183, opened 2025-03-21
Don't add an Authorization header is there is no token

### LINKED ISSUE #182: `TokenAuth.get_auth_header` returns invalid data when token is `None`
When connecting to a registry for the first time, without a bearer token, the following Authorization header is sent : `Bearer None`. It seems to break some registry ( Jfrog's Artifactory in our case ) and make them reply with and invalid auth header.

I checked and neither docker, cri-o, podman nor flux sends an empty authorization header.

Meallia:
fixes #182


vsoch:
Looks good! You'll need to run black, bump the version in the CHANGELOG and change in oras/version.py.

tarilabs:
thank you @Meallia for the fix and improvement, possibly with the more recent refactoring this is getting invoked when not necessary, either way lgtm as both

vsoch (approved):

PR #186, opened 2025-03-31
Propagate the tls_verify parameter to the auth backends

Meallia:
In the current version, when using a registry / client with tls_verify set to False on a registry requiring authentication, the client will fail with an SSL error.

When creating the Auth backend, the tls_verify parameter was not used.

While testing this, I ran into issues with the `retry` decorator waiting for SSL errors to magically disappear so I added an exception for those. Let me know if you'd rather i removed it.


oras/auth/base.py
```diff
@@ -17,6 +19,9 @@ class AuthBackend:
     Generic (and default) auth backend.
     """
 
+    session: requests.Session
+    _tls_verify: bool
```
vsoch, line 23: Why would you make the tls verify a class variable? This means multiple classes must have the same state, and if you change one the others follow.

vsoch (commented):
Great job on the tests! I just had one question, and then please update the CHANGELOG.md as well.

oras/auth/base.py
vsoch: The base class [does have an init](https://github.com/oras-project/oras-py/blob/938390ae8d283ded95924c9fc87ffeea53d38e2d/oras/auth/base.py#L20) - it primarily doesn't include many variables that were added after.

vsoch (approved):
Thanks for adding this @Meallia.

PR #190, opened 2025-04-24
Avoid Setting Unspecified Header Fields When Pushing Manifests

Sojamann:
This PR addresses compatibility issues with various OCI registry implementations. Some registries are more strict than others when it comes to certain HTTP headers, which can lead to unexpected errors or behavior.

Aside from the `Content-Length` being calculated incorrectly, the [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests) does not require or even mention this header when pushing manifests.

Let me know if I overlooked some detail and thanks for your work ✌️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - don't include Content-Length header in upload_manifest (0.2.29)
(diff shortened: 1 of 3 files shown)


vsoch:
This should be OK to do - it looks like content length isn't stated in the spec for pushing a manifest:

https://github.com/opencontainers/distribution-spec/blob/583e014d15418d839d67f68152bc2c83821770e0/spec.md#pushing-manifests

Could you please sign the DCO, and then bump the verison in version.py and add a note with the corresponding version and change to the CHANGELOG.md?

Sojamann:
done

vsoch (approved):

Issue #185, opened 2025-03-24
Client certificate/private key file support


mawiegand:
Are there plans to support client certificates/private key files for the remote registry?

mawiegand:
Yes, it is supported. See `--cert-file` and `--key-file` on https://oras.land/docs/commands/oras_pull#options

vsoch:
I'd be happy to review a PR that adds this functionality then.

PR #197, opened 2025-05-02
Add TLS Validation

Sojamann:
This PR adds tls verification options and solves #185.

---
At the moment only _insecure_ (no tls verification) is supported which I'd like to change since I cannot rely on the hosts CA Bundles which _requests_ uses by [default](https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification).

The tls options are also supported by the [oras binary](https://oras.land/docs/commands/oras_blob_push#options).

### DIFF
--- a/oras/auth/__init__.py
+++ b/oras/auth/__init__.py
@@ -14,14 +14,11 @@ class AuthenticationException(Exception):
     pass
 
 
-def get_auth_backend(
-    name="token", session=None, insecure=False, tls_verify=True, **kwargs
-):
+def get_auth_backend(name="token", session=None, insecure=False, **kwargs):
     backend = auth_backends.get(name)
     if not backend:
         raise ValueError(f"Authentication backend {backend} is not known.")
     backend = backend(**kwargs)
     backend.session = session or requests.Session()
     backend.prefix = "http" if insecure else "https"
-    backend._tls_verify = tls_verify
     return backend
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - allow specification of server and client certificates (0.3.0)
  - initialize headers variable in do_request (0.2.31)
  - Make reproducible targz without mimetype (0.2.30)
  - don't include Content-Length header in upload_manifest (0.2.29)
--- a/oras/auth/base.py
+++ b/oras/auth/base.py
@@ -20,7 +20,6 @@ class AuthBackend:
     """
 
     session: requests.Session
-    _tls_verify: bool
 
     def __init__(self, *args, **kwargs):
         self._auths: dict = {}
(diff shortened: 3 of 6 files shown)


oras/auth/__init__.py
```diff
@@ -14,14 +14,11 @@ class AuthenticationException(Exception):
...
 
-def get_auth_backend(
-    name="token", session=None, insecure=False, tls_verify=True, **kwargs
-):
+def get_auth_backend(name="token", session=None, insecure=False, **kwargs):
     backend = auth_backends.get(name)
     if not backend:
         raise ValueError(f"Authentication backend {backend} is not known.")
     backend = backend(**kwargs)
     backend.session = session or requests.Session()
```
Sojamann, line 22: This kinda is a problem if someone does not pass in the _session_... Is this something that needs to be taken into account or can we not just make _session_ a required argument here. Having ONE session which is pre-configured to use for all requests has the advantage that we can pass less arguments around settings like 

```py
self.session.cookies.set_policy(DefaultCookiePolicy(allowed_domains=[]))
```
cannot be forgotten

Sojamann:
The test is failing since I removed the explicit  `verify=False` setting in each request, which leads to the issue described [here](https://github.com/psf/requests/issues/3829).

The session has `verify=False` but when the  _verify_ parameter is not explicitly set in the request, it will default to checking the env variable _REQUESTS_CA_BUNDLE_ which takes precedence over the session setting.

You might have known this, I just learned this today  ... 😲

Since the user can now set the CA bundle explicitly, would it make sense to set `trust_env = False` on the session to prevent it from reading environment variables? This would make the code also less dependent on the underlying request library.

What are your thoughts? How should we proceed?

oras/auth/__init__.py
vsoch: The one session and passing it forward from the client was the idea - see how it is created and passed forward here: https://github.com/oras-project/oras-py/blob/9e29e24ac4948c8efc1756e8bb631c6149c4d45f/oras/provider.py#L68

vsoch:
If the session is shared, for your implementation can you just create the client and set session.verify = `/path/to/bundle` instead?

oras/auth/__init__.py
Sojamann: should the function `get_auth_backend` stay like this since it is correctly used for should session be made a required parameter?

vsoch: Sorry I don't understand - do you mean _or_ should session be made required?

Sojamann:
@vsoch do you mean that this change should not be made and the user of the `Registry` should set the cert like this?

```py
r = Registry(...)
r.session.verify = '/path/to/my/bundle'
```

oras/auth/__init__.py
Sojamann: Either we leave it like it is know and it works fine the way it is currently used or we change it to

```py
def get_auth_backend(session: requests.Session, name="token", , insecure=False, **kwargs):
```

so that no that there can no mistakes be made where in the future somebody does not pass in the session and just does:

```py
get_auth_backend(name="token", insecure=False)
```

vsoch:
@Sojamann yes I'm wondering if you'd like to try that and see if it works for your use case.

Sojamann:
This would also work, but as a user of Oras, I would then be exposed to the internals of the library.

vsoch:
@Sojamann it's a developer library by default - I would argue that your use case is niche and should not refactor the entire structure of the library to support when it's as easy as tweaking the session on the class.

Sojamann:
Hm I’m not fully on board with this, but I guess it is what it is. I’ll go ahead and close the PR then.

PR #200, opened 2025-05-26
fix(auth): call 'load_configs()' to load auth infos from '.docker/con…

diverger:
In provider.py, both 'push()' and 'pull()' call 'load_configs()'. There is no reason 'get_manifest()' won't call it. 

One case is if you call 'docker/login-action' to login the GHCR, it will save the auth infos in '.docker/config.json', without calling 'load_configs()', it can't load the user/password and will throw:

`'TokenAuth' object has no attribute '_basic_auth'`

### DIFF
--- a/oras/auth/token.py
+++ b/oras/auth/token.py
@@ -124,10 +124,18 @@ class TokenAuth(AuthBackend):
             logger.debug(f"Scope: {h.scope}")
             params["scope"] = h.scope
 
-        # Set Basic Auth to receive token
-        headers["Authorization"] = "Basic %s" % self._basic_auth
+        # Set Basic Auth to receive token, if available
+        if hasattr(self, "_basic_auth") and self._basic_auth:
+            headers["Authorization"] = "Basic %s" % self._basic_auth
+            logger.debug("Using Basic Auth for token request.")
+        else:
+            logger.debug(
+                "No Basic Auth available or configured for token request. Proceeding without Basic Auth header for token endpoint."
+            )
 
-        logger.debug(f"Requesting auth token for: {h}")
+        logger.debug(
+            f"Requesting auth token for: {h} with header keys: {list(headers.keys())}"
+        )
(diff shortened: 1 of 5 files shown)


oras/auth/token.py
```diff
@@ -124,10 +124,14 @@ def request_token(self, h: auth_utils.authHeader) -> bool:
             logger.debug(f"Scope: {h.scope}")
             params["scope"] = h.scope
 
-        # Set Basic Auth to receive token
-        headers["Authorization"] = "Basic %s" % self._basic_auth
-
-        logger.debug(f"Requesting auth token for: {h}")
+        # Set Basic Auth to receive token, if available
```
vsoch, line 127: It seems like this is mixing up the token and basic auth modules?

vsoch: Gotcha. @diverger you will need to sign the DCO and update the version in version.py and make a corresponding entry in the CHANGELOG.

diverger: @vsoch Done it.

diverger: @vsoch It seems the reformatting of token.py cause the merge failure?

vsoch: Yes, you'll need to install the same version of black that we have pinned and run again.

diverger: @vsoch token.py has been formatted.

vsoch (approved):
Thanks!

diverger:
It seems some other operations less this operation too.


________________________________
发件人: Vanessasaurus ***@***.***>
已发送: 2025 年 5 月 31 日 星期六 8:13
收件人: oras-project/oras-py ***@***.***>
抄送: diverger ***@***.***>; Mention ***@***.***>
主题: Re: [oras-project/oras-py] fix(auth): call 'load_configs()' to load auth infos from '.docker/con… (PR #200)


Merged #200<https://github.com/oras-project/oras-py/pull/200> into main.

―

Issue #202, opened 2025-05-27
Issue accessing registry.redhat.io with ~/.docker/config.json


benruland:
I am having difficulties accessing registry.redhat.io with a provided ~/.docker/config.json.

The error is:

```
AttributeError: 'TokenAuth' object has no attribute '_basic_auth'
```

However, it feels like oras is not reading my config.json

Been reading through the code, but I can not comprehend the logic, how oras-py reads the config.json and how to debug that.

## Steps to reproduce

I am running in a container with Python 3.9.21 and oras==0.2.31.

In python, I can reproduce it with these commands and a `~/.docker/config.json`

```
from oras.client import OrasClient
import logging
logging.basicConfig(
    format="%(asctime)s %(levelname)-8s %(message)s",
    level="DEBUG",
    datefmt="%Y-%m-%d %H:%M:%S",
    encoding="utf-8",
)
oras_client = OrasClient()
print(oras_client.get_tags("registry.redhat.io/openshift-logging/eventrouter-rhel8"))
```

## Details for my environment and error

```
# Token from https://access.redhat.com/terms-based-registry/accounts
$ cat ~/.docker/config.json
{"auths":{"quay.io":{"auth":"REDACTED=="},"registry.redhat.io":{"auth":"REDACTED"},"registry.connect.redhat.com":{"auth":"REDACTED=="}}}

$ (app-root) sh-5.1$ pip freeze
atlassian-python-api==3.41.21
attrs==25.3.0
beautifulsoup4==4.13.4
build==1.2.2.post1
certifi==2025.4.26
charset-normalizer==3.4.2
Deprecated==1.2.18
gitdb==4.0.12
GitPython==3.1.44
hvac==2.3.0
idna==3.10
importlib_metadata==8.7.0
jmespath==1.0.1
jsonschema==4.24.0
jsonschema-specifications==2025.4.1
oauthlib==3.2.2
openshift-client==2.0.5
oras==0.2.31
packaging==25.0
pyproject_hooks==1.2.0
python-gitlab==5.6.0
PyYAML==6.0.2
referencing==0.36.2
requests==2.32.3
requests-oauthlib==2.0.0
requests-toolbelt==1.0.0
rpds-py==0.25.1
six==1.17.0
smmap==5.0.2
soupsieve==2.7
tomli==2.2.1
typing_extensions==4.13.2
urllib3==2.4.0
wrapt==1.17.2
zipp==3.22.0

(app-root) sh-5.1$ python
[...]

PR #204, opened 2025-05-31
Load auth configs before making requests to container registry

diverger:
After fix issue #200, I see someone encountered similar problem like #202, and I dive further into the code, and find more methods lack the calling of 'load_configs()'.


vsoch:
Why are you calling this in one off functions instead of at the class init?

diverger:
Because the 'container' passed in as an argument which can only be determined at runtime. And calling 'load_configs()' align with the behavior of push()/pull() which have been calling it already. Yes, it will be more elegant to make the duplicated calls to some re-usable code. But that may need more changes and may break something 😀.

vsoch:
The docker config is not going to change based on the container. If a user needs to login to registries they should do that before using the tool. Please have the call done once, here: 

https://github.com/oras-project/oras-py/blob/1861cb655e16a3697e4d50baa2107a65ec0d80b7/oras/provider.py#L86

on init.

diverger:
Yes, if the strategy is to only allow the client to initialize authorization-related settings during initialization (not obtained from container address in methods calling), init is a suitable place. But we might need to consider how to handle the “hostname”. Load the related configs if provided? But what if it's omitted?

vsoch:
We already load all the auths. https://github.com/oras-project/oras-py/blob/b63ede0287865d5884812312f7e90ca44724296f/oras/auth/utils.py#L36. And it's just the case that one is chosen based on a specific container registry (the hostname). Instead we should have that functionality to update the auth (and change) to coincide with any function that requires (and loads a container via that decorator). E.g.,

```bash
 @decorator.ensure_container
 @decorator.ensure_auth
    def delete_tag(self, container: container_type, tag: str) -> bool:
```
Then we aren't unnecessarily re-reading the user docker config. If we know ensure_container is going to add the container, then you could add another decorator to ensure_auth, and you would want to make sure the correct credential is just available to use (but not force it to be used until a prompt from the registry).

diverger:
@vsoch Oh, I didn't realize the presence of auths.update(cfg.get("auths", {})). If the authorization-related information is already there, then delay loading the authorization as needed makes perfect sense.

diverger:
@vsoch The current 'blob_exists()' use 'layer: dict' as first arg, and 'container' as second, so the test fail. It seems it's not aligned with 'ensure_container()' or some other methods, such as 'put_upload'.

vsoch:
I would guess it is happening because target is being provided as a kwarg, and in that case, the decorator looks for "container."

vsoch:
The call that triggers the error doesn't provide args, it provides the key word arguments "kwargs" because it explicitly says target=.

diverger:
@vsoch please review it.

diverger:
Do you mean to use one decorator to handle both

```
    def upload_manifest(
        self,
        manifest: dict,
        container: oras.container.Container,
    ) -> requests.Response:
```

And

```
    def download_blob(
        self, container: container_type, digest: str, outfile: str
    ) -> str:
```

Without breaking the users' current code?

diverger:
@vsoch The new commit uses a more 'unified' style, but the underlying logic is the same with the 'ensure_auth_second_arg'.

vsoch:
> Do you mean to use one decorator to handle both

Yes, if we already have logic to handle the container arg, you shouldn't need to add new decorators or functions. I didn't consider this approach at the beginning in the case we would want to separate those two things, but I don't see if likely that we need to.

PR #206, opened 2025-06-07
Add support for Docker credsStore and credHelpers

rasmusfaber:
This updates the basic backend to also support using Docker credsStore and Docker credHelpers.

```python
client = oras.client.OrasClient(auth_backend="basic")
client.pull(target="123456789012.dkr.ecr.us-east-1.amazonaws.com/repo:latest")
```

### DIFF
--- a/oras/auth/base.py
+++ b/oras/auth/base.py
@@ -2,7 +2,8 @@ __author__ = "Vanessa Sochat"
 __copyright__ = "Copyright The ORAS Authors."
 __license__ = "Apache-2.0"
 
-
+import json
+import subprocess
 from typing import Optional
 
 import requests
@@ -23,7 +24,7 @@ class AuthBackend:
     _tls_verify: bool
 
     def __init__(self, *args, **kwargs):
-        self._auths: dict = {}
+        self._auth_config: dict = {}
         self.prefix: str = "https"
 
     def get_auth_header(self):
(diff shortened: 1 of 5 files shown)


oras/auth/base.py
```diff
@@ -48,20 +49,41 @@ def logout(self, hostname: str):
...
                 logger.info(f"You have successfully logged out of {hostname}")
                 return
         logger.info(f"You are not logged in to {hostname}")
 
     def _logout(self):
         pass
 
+    def _get_auth_from_creds_store(self, binary: str, hostname: str) -> str:
+        try:
+            proc = subprocess.run(
```
vsoch, line 69: Is this really best practice for doing this? it feels a bit scary to be able to execute an arbitrary binary. That said, the docker pycreds library that provides equivalent functionality does similar:

https://github.com/shin-/dockerpy-creds/blob/9c0b66d2e445a838e1518f2c3273df7ddc7ec0d4/dockerpycreds/store.py#L3

vsoch: And I like the added functionality. I suspect many folks use these credsHelpers.

rasmusfaber: > Is this really best practice for doing this? it feels a bit scary to be able to execute an arbitrary binary. That said, the docker pycreds library that provides equivalent functionality does similar:
> 
> https://github.com/shin-/dockerpy-creds/blob/9c0b66d2e445a838e1518f2c3273df7ddc7ec0d4/dockerpycreds/store.py#L3

Yes, I think it is the standard way to use these. And the docker cli will do the same: https://github.com/docker/docker-credential-helpers/blob/master/client/client.go#L46-L73 .

rasmusfaber: > And I like the added functionality. I suspect many folks use these credsHelpers.

At least it is an extremely flexible way of allowing credentials to be configured. #205 works for AWS ECR, but this should work for basically everything (Azure, GCR, etc.).

vsoch (approved):

oras/auth/base.py
```diff
@@ -70,21 +92,35 @@ def _load_auth(self, hostname: str) -> bool:
...
+                # no auth there (wonky file)
                 return False
             self._basic_auth = auth
             return True
+        # Check for credsStore:
+        if self._auth_config.get("credsStore"):
+            auth = self._get_auth_from_creds_store(
+                self._auth_config["credsStore"], hostname
+            )
+            if auth is not None:
```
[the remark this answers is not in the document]
vsoch: That function either raises an error or returns:

```python
return auth_utils.get_basic_auth(payload["Username"], payload["Secret"])
```
And that does always return a dictionary of values. I don't think it's a bug (at least superficially) it's more just an unnecessary check.

PR #213, opened 2025-06-18
fix Python 3.7+ compatibility by avoiding union syntax and rely on typing

tarilabs:
Oras is declared as Python 3.7+ compatible:

https://github.com/oras-project/oras-py/blob/b648fd36cb600537d33597e4c2291575770283bc/setup.py#L105

but #212 introduced newer Python syntax.

Fix propose to rely on `typing`.

cc @rasmusfaber @vsoch


rasmusfaber (approved):
Looks good. Thanks for fixing this.

PR #218, opened 2025-09-10
Fix: runtime error can not be emitted successfully

kerthcet:
Happened when try to delete the blob with error configuration.


vsoch (commented):
Good catch - thank you! Before we merge - does a non 202 response also correctly format to json?

kerthcet:
any feedback? @vsoch

Issue #108, opened 2023-10-05
Image Index support


miker985:
As of 0.1.25 it doesn't appear to me this supports Image Indexes.

The lift looks small - maybe as little as adding a media type to defaults.py, a schemas.py definition, and minor work to support it in the Registry object

Sample code might be

```python
class Registry(oras.provider.Registry):
    """
    Oras registry with support for image indexes.
    """

    @decorator.ensure_container
    def get_image_index(self, container, allowed_media_type=None):
        """
        Get an image index as a manifest.

        This is basically Registry.get_manifest with the following changes

        - different default allowed_media_type
        - no JSON schema validation
        """
        if not allowed_media_type:
            default_image_index_media_type = "application/vnd.oci.image.index.v1+json"  # TODO: need this in defaults.py
            allowed_media_type = [default_image_index_media_type]

        headers = {"Accept": ";".join(allowed_media_type)}

        manifest_url = f"{self.prefix}://{container.manifest_url()}"
        response = self.do_request(manifest_url, "GET", headers=headers)
        self._check_200_response(response)
        manifest = response.json()
        # TODO: jsonschema.validate(manifest, schema=...)
        return manifest
```

If there's interest please let me know and I can put in a PR. In that case I ask if a new method like `get_image_index` is desired or if something else would be preferred.

In any case, thank you for the project!

vsoch:
Perhaps under examples https://github.com/oras-project/oras-py/tree/main/examples would suffice? Can you give me the use case of doing this?

miker985:
My use case is downloading an image index, selecting a platform-compatible manifest, and then following that to ultimately download a platform-specific artifact.

vsoch:
That would be a great (full) example script if you are up for it!

PR #237, opened 2026-02-17
Use optional schema validation for `get_manifest`

ianpittwood:
- Validate artifact retrieved by `get_manifest` only if a schema is passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
+ - use dynamic validation of response contents for `get_manifest()` instead of always expecting a manifest (0.2.41)
  - fix preemptive exit in non-empty `auths` lookup when `credsStore` or `credHelpers` is used (0.2.40)
  - introduce API to copy from oci-layout to OCI registry (0.2.39)
  - fix Python 3.7+ compatibility by avoiding union syntax and rely on typing (0.2.38)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
- - use dynamic validation of response contents for `get_manifest()` instead of always expecting a manifest (0.2.41)
+ - make `get_manifest()` validation optional, fix `Accept` header join, and expand default `Accept` header types to cover all supported response types for the `/v2/<name>/manifests/<reference>` endpoint (0.2.41)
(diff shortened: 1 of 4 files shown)


ianpittwood:
@vsoch I had this as an idea to tackle partial Index support per https://github.com/oras-project/oras-py/issues/108. 

My logic is that the `/v2/<name>/manifests/<reference>` endpoint can return either the `application/vnd.oci.image.manifest.v1+json` type or `application/vnd.oci.image.index.v1+json` and therefore the validation should not be hardcoded to only validate against the manifest schema. I don't think it makes sense to have separate methods either when an identical `GET` request is being made.

I may have gone a bit overboard. I thought it would be handy to have it automatically validate the appropriate schema for the content in the response rather than having to explicitly provide it. TIA, let me know what you think!

vsoch:
This isn't a bad idea, although I wonder what cases validation would not be desired. What does oras (in Go) do?

ianpittwood:
Looks like it [accepts all compatible content types for the manifest endpoint](https://github.com/oras-project/oras-go/blob/42c1a3adcfba8ae373e3fe53308867f0c7357dc2/registry/remote/manifest.go#L50) and then it [treats all references as generic OCI descriptors](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L1199) before later differentiation and implicit type checking based on the content media type.

Then for index handling, I think it follows the similar DAG resolution logic used to pull each blob that composes a manifest. There's just another layer added on top to resolve the index to each manifest to each blob. It doesn't seem like it would be too terrible to implement into the existing pull logic.

vsoch:
If I understand correctly - they don't do validation by default.

I think we should preserve that here, and only validate explicitly if the user requests it via the get manifest call.

oras/provider.py
```diff
@@ -945,21 +946,25 @@ def get_manifest(
         :type container: oras.container.Container or str
         :param allowed_media_type: one or more allowed media types
         :type allowed_media_type: str
+        :param validation_schema: optional json schema to validate the manifest against, defaults to media type if a
```
vsoch: I think we just want to say it's an optional json schema. If it is not provided, it just defaults to None.
```suggestion
        :param validation_schema: optional json schema to validate the manifest against
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
```
vsoch, line 16: And probably should update the note here too. Then I think we will be good!

ianpittwood:
Sorry, pushed before I was actually done updating!

I think the only common elements that we could always validate are the `schemaVersion` which should always be `2` and the `mediaType` which should always be one of the four media types supported for the `manifests` endpoint. Beyond that it seems that each type is at least slightly different with its expected fields. I think that's the only validation that could be done reliably at `get_manifest`'s place in the workflow.

I switched `get_manifests` to only do schema validation if a value is passed to the `validation_schema` kwarg. I also noticed that the `Accept` header was being joined using `;` rather than `,` which is what should be used to join different media types. I also expanded the default accepted media types to match those used in `oras-go`, which covers all the possible media types that can be returned by the `manifests` endpoint.

vsoch (approved):

PR #238, opened 2026-02-21
feat: add Layout `copy` for pull_from_registry capability

tarilabs:
- follow up on previous PR for copy, now allow Pull (from Registry to OCI Layout)
  - https://github.com/oras-project/oras-py/pull/228
- re-use code and code-conventions
  - made sure I didn't leave any single-use variables (as requested previously)
  - re-used utilities for json serde
  - re-used naturally what was made previously in https://github.com/oras-project/oras-py/pull/228 as possible
  - reason for parameter called `pull_from_registry(...  target: str, ...)` is to make it consistent with [analogous use in provider](https://github.com/oras-project/oras-py/blob/e51bc875f8eee6a120e7fb873362439194ec2eb4/oras/provider.py#L871)
- introduced `NewLayoutFromRegistry` convenience/factory method (as analogously requested previously)
- naturally test coverage, re-using fixture and style as possible

Hope this is helpful! 🙏 🚀 

cc @vsoch  @astefanutti

### DIFF
--- a/oras/layout/layout.py
+++ b/oras/layout/layout.py
@@ -19,7 +19,7 @@ from oras.layout.validation import (
     _validate_oci_layout_file,
 )
 from oras.logger import logger
-from oras.utils.fileio import read_json
+from oras.utils.fileio import read_json, write_json
 
 if TYPE_CHECKING:
     from oras.provider import Registry
@@ -39,6 +39,45 @@ def NewLayout(path: str, validate: bool = True) -> Layout:
     :raises ValueError: if path is not a directory, or validation fails (when validate=True)
     """
     return Layout(path=path, validate=validate)
+
+
+def NewLayoutFromRegistry(
+    path: str,
+    provider: Registry,
+    target: str,
+    tag: str = "latest",
+) -> Layout:
+    """
+    Create a new OCI Layout on disk by pulling from a remote registry.
+
+    The path must either not exist or be an empty directory.
+
+    :param path: path to create the Layout directory
+    :type path: str
+    :param provider: Registry provider instance for downloading
+    :type provider: oras.provider.Registry
+    :param target: source registry/repository with tag or digest
+        (e.g., "ghcr.io/user/repo:v1.0" or "ghcr.io/user/repo@sha256:abc123...")
+    :type target: str
+    :param tag: tag to write in the layout's index.json annotation (default: "latest")
+    :type tag: str
+    :return: Layout instance for the pulled Layout directory
+    :rtype: Layout
+    :raises FileExistsError: if path exists and is not an empty directory
+    :raises ValueError: if path exists but is not a directory
+    """
+    p = pathlib.Path(path)
+    if p.exists():
+        if not p.is_dir():
+            raise ValueError(f"Path exists and is not a directory: {path}")
+        if any(p.iterdir()):
+            raise FileExistsError(f"Directory is not empty: {path}")
+    else:
+        p.mkdir(parents=True)
+
+    layout = Layout(path=path, validate=False)
+    layout.pull_from_registry(provider=provider, target=target, tag=tag)
+    return layout
 
 
 class Layout:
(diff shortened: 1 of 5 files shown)


oras/tests/test_layout.py
```diff
@@ -572,6 +572,37 @@ def test_get_ordered_blobs_tag_not_found():
         Layout(layout_path).get_ordered_blobs("nonexistent")
 
 
+def test_blob_exists():
+    """Test _blob_exists returns True for blobs on disk and False otherwise"""
+    layout_path = pathlib.Path(__file__).parent / "ocilayout_data/ocilayout1"
+
+    # known blobs from ocilayout1
+    assert (
```
vsoch, line 580: For the way you are using this class function, should `_blob_exists` be public and not private?

vsoch: And what is the benefit or use case from doing Layout.blob_exists as a class function vs. creating the layout first then having an instance function? E.g.,

```
layout = NewLayoutFromRegistry(path)
layout.blob_exists("shaxxxxxxx")
```

oras/tests/test_layout.py
```diff
@@ -672,3 +703,214 @@ def test_push_from_layout_invalid_tag(
...
+
+    pulled_layout.validate()
+    source_blobs = sorted(Layout(source_layout_path).get_ordered_blobs("latest"))
+    pulled_blobs = sorted(Layout(pull_dir).get_ordered_blobs("latest"))
+    assert source_blobs == pulled_blobs
+
+    # compare test oci-layout with Pulled tmp oci-layout bytewise
+    for digest in source_blobs:
+        source_path = Layout._digest_to_blob_path(
+            pathlib.Path(source_layout_path), digest
```
vsoch: Perhaps wrapping in pathlib.Path should be handled in this function for the user, if not done? And any reason again for this to be private if you are using it outside of the class like this?

oras/tests/test_layout.py
tarilabs: done with 470c6e6

oras/tests/test_layout.py
tarilabs: done with 470c6e6

tarilabs (commented):
thanks @vsoch , I've made both methods instance methods and public now.

Therefore also made `_process_manifest`, `_pull_manifest_blobs`, `_pull_index_blobs` instance method as well, so to avoid passing around the layout path param. 

There was also one chance of a small refactor which I did too, as I didn't realize I did have a small duplicated code lingering around previously.

oras/layout/layout.py
```diff
@@ -347,3 +518,100 @@ def push_to_registry(
...
+
+        # prepare directory structure for blobs
+        blobs_dir = layout_dir / oras.defaults.oci_blobs_dir / "sha256"
+        blobs_dir.mkdir(parents=True, exist_ok=True)
+
+        headers = {
+            "Accept": ", ".join(oras.defaults.default_manifest_accepted_media_types)
+        }
+        manifest_url = f"{provider.prefix}://{container.manifest_url()}"
+        response = provider.do_request(manifest_url, "GET", headers=headers)
```
vsoch, line 556: In the future we should take a look at this "check 200 response" function. I'm wondering if there are ever cases we wouldn't want to check for a specific response. If we always want to check, we might want to integrate that into the do_request and have the check done before return.

Not needed for this PR, just thinking out loud.

vsoch (commented):
Looks great! Just the changelog and version and we are good to merge.

oras/layout/layout.py
tarilabs: Noted as https://github.com/oras-project/oras-py/issues/239

tarilabs:
> Looks great! Just the changelog and version and we are good to merge.

Thank you so much!
fwiw I will be on mobile only for some days: feel free to directly append on this PR to merge it sooner (Maintainer rights to the pr branch are kept), or I will do it on my return!

astefanutti (approved):
Thanks @tarilabs!

tarilabs:
added changelog and version, rebased and squashed af1aadd , 470c6e6

vsoch (approved):

PR #243, opened 2026-08-01
Add `Registry.resolve_digest()`

### LINKED ISSUE #143: Add `get_digest(container)` method on provider.Registry
It can often be quite useful to resolve a tag to the digest where an artifact is stored in the registry. To solve this I duplicated the code of `Registry.get_manifest` and compute the hash client side:
```python
    @decorator.ensure_container
    def get_digest(self, container):
        self.load_configs(container)
        allowed_media_type = [defaults.default_manifest_media_type]
        headers = {"Accept": ";".join(allowed_media_type)}
        headers.update(self.headers)
        get_manifest = f"{self.prefix}://{container.manifest_url()}"  # type: ignore
        response = self.do_request(get_manifest, "GET", headers=headers)
        self._check_200_response(response)
        return f"sha256:{hashlib.sha256(response.content).hexdigest()}"
```

judahrand:
Resolves https://github.com/oras-project/oras-py/issues/143

### DIFF
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -3,6 +3,7 @@ __copyright__ = "Copyright The ORAS Authors."
 __license__ = "Apache-2.0"
 
 import copy
+import hashlib
 import os
 import sys
 import urllib
@@ -965,6 +966,46 @@ class Registry:
         if validation_schema:
             jsonschema.validate(manifest, schema=validation_schema)
         return manifest
+
+    @decorator.ensure_container
+    def resolve_digest(
+        self,
+        container: container_type,
+        allowed_media_type: Optional[List[str]] = None,
+    ) -> str:
+        """
+        Resolve a manifest reference to its digest.
+
+        :param container: parsed container URI
+        :type container: oras.container.Container or str
+        :return: canonical manifest digest, including the algorithm prefix
+        :param allowed_media_type: one or more allowed media types
+        :type allowed_media_type: list or None
+        :rtype: str
+        """
+        # Load authentication configs for the container's registry
+        # This ensures credentials are available for authenticated registries
+        self.auth.load_configs(container)
+
+        if not allowed_media_type:
+            allowed_media_type = oras.defaults.default_manifest_accepted_media_types
+        headers = {"Accept": ", ".join(allowed_media_type)}
+        manifest_url = f"{self.prefix}://{container.manifest_url()}"  # type: ignore
+
+        response = self.do_request(manifest_url, "HEAD", headers=headers)
+        if response.status_code not in [405, 501]:
+            self._check_200_response(response)
+            digest = response.headers.get("Docker-Content-Digest")
+            if digest:
+                return digest
+
+        # Older registries may not support HEAD or provide a digest header.
+        response = self.do_request(manifest_url, "GET", headers=headers)
+        self._check_200_response(response)
+        digest = response.headers.get("Docker-Content-Digest")
+        if digest:
+            return digest
+        return f"sha256:{hashlib.sha256(response.content).hexdigest()}"
 
     @decorator.retry()
     def do_request(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/oras-project/oras-py/tree/main) (0.0.x)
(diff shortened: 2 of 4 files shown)


oras/provider.py
```diff
@@ -966,6 +967,46 @@ def get_manifest(
...
+        manifest_url = f"{self.prefix}://{container.manifest_url()}"  # type: ignore
+
+        response = self.do_request(manifest_url, "HEAD", headers=headers)
+        if response.status_code not in [405, 501]:
+            self._check_200_response(response)
+            digest = response.headers.get("Docker-Content-Digest")
+            if digest:
+                return digest
+
+        # Older registries may not support HEAD or provide a digest header.
```
vsoch, line 1002: Just a quick question here. So we are comparing a digest we just got from the header with... the same request via a head to get the digest? My instinct is that we would want to check the content against the digest we received. It's unlikely doing two calls to get an equivalent digest would be different. Let me know if I'm not understanding this function. Thank you!

judahrand: The first call is HEAD rather than GET and so doesn't return the manifest. The OCI Distribution Spec says that it should return the digest header though.

So in the first instance we do a HEAD and check for the header. If the header is present we just return the digest directly (line 1000).

If the header isn't present or the registry doesn't support HEAD twe do the full GET.

If the registry previously didn't support HEAD at all it still might give us the digest header so if it does we return that directly (line 1007).

Only if the GET also doesn't give use the header do we calculate it ourselves (line 1008).

This is an optimisation to save network bandwidth and CPU time calculating the hash in the happy path. We don't compare what is in the header to the content at any point - we could do this though if you think it is worthwhile. I've made the assumption that the registry can be trusted to give the correct information in the header as the spec requires.

vsoch: Ah ok, I think I was caught up in the test, which is indeed checking something we just derived:

```python
digest = res.headers["Docker-Content-Digest"]
assert remote.resolve_digest(target) == digest
```

We might also consider adding:

```console
manifest_bytes = <raw GET of the manifest>
assert remote.resolve_digest(target) == f"sha256:{hashlib.sha256(manifest_bytes).hexdigest()}"
```

In practice (outside of testing) the user would have a target URI that they want to verify against the registry.  A quick question - what happens if we get a target with a sha in it? E.g., `resolve_digest(<container>@sha256:abc")` Do we need to validate the format of the docker digest (do registries vary)?

judahrand: > We might also consider adding:

I'll add this when at my keyboard. 

> A quick question - what happens if we get a target with a sha in it? 

Hmm... Good question. Currently we should end up returning a digest which maps to the same artifact (though it might use a different hash). Do you think it would be better to just parse out the digest and return it to the caller without speaking to the registry?

> Do we need to validate the format of the docker digest (do registries vary)?

I suppose that depends on how much we trust the registry. No big downside to validating it, however, given that the spec does layout a regex format: https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests

vsoch: > I'll add this when at my keyboard.

Thank you!

> Hmm... Good question. Currently we should end up returning a digest which maps to the same artifact (though it might use a different hash). Do you think it would be better to just parse out the digest and return it to the caller without speaking to the registry?

Let's have it go through the operation now. At first I was thinking maybe, but I'm worried if that could be used in some malicious way. 

> I suppose that depends on how much we trust the registry. No big downside to validating it, however, given that the spec does layout a regex format: https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests

I'm OK to be conservative and not add some validation until we have reason to.

judahrand: >We might also consider adding:
> 
> manifest_bytes = <raw GET of the manifest>
> assert remote.resolve_digest(target) == f"sha256:{hashlib.sha256(manifest_bytes).hexdigest()}"

I just took another look and I don't think we really need this.

From the existing lines you picked out:

```python
digest = res.headers["Docker-Content-Digest"]
assert remote.resolve_digest(target) == digest
```

The `digest` is the header returned by the registry when we pushed the artifact. This is defined as the cannonical digest by the spec. We're then testing that `resolve_digest` returns the same, cannonical digest. 

If might be worth testing the other code paths with unittests and mocks? But I don't see that done for any of the other code really.

vsoch: Would not hurt to start! :wink:

judahrand: Added some unittests

vsoch:
@FeynmanZhou I think we need another review and also to clarify what the LLM usage policy is for adding content generated by an LLM.

PR #244, opened 2026-08-02
Fix issues with temporary files leaking

### LINKED ISSUE #246: Pulling an artifact with media type `application/vnd.oci.image.layer.v1.tar+gzip` leaks a temporary file
The temporary file created [here](https://github.com/oras-project/oras-py/blob/a209c6cc145869e3056a7061398efca1c14f39d7/oras/provider.py#L922) is never removed. In long living applications this can result in excessive disk usage or memory usage depending on how the underlying filesystem is implemented.

judahrand:
There are currently some issues with temporary files not being removed. This is particularly notable in the pull codepath - if  we're pulling a tar it is downloaded to a tmpfile and then extracted to the final destination but the tmpfile is never cleaned up.

This PR refactors the temp file helpers in `utils` to be context managers in order to help force deletion. The tempdir created in the pull path is still not deleted but that is clearly by design.

This change was LLM assisted but by no means "vibe-coded". If ORAS has an LLM policy I'm happily to adhere but I can't find one!

Resolves #246


judahrand:
I can implement a simpler fix if desired. The larger refactor was motivated by trying to ensure the same problems don't get reimplemented.

PR #245, opened 2026-08-02
Add content verification

### LINKED ISSUE #247: Project does not verify that the content provided by a remote registry matches what was requested
The [OCI Distribution Spec 1.1.0](https://specs.opencontainers.org/distribution-spec/?v=v1.1.0#content-verification) gives guidance that consumers should verify content from an untrusted registry.

The spec also says:

> The Docker-Content-Digest header, if present on the response, returns the canonical digest of the uploaded blob which MAY differ from the provided digest. If the digest does differ, it MAY be the case that the hashing algorithms used do not match. See [Content Digests](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests) [apdx-3](https://specs.opencontainers.org/distribution-spec/?v=v1.1.1#appendix) for information on how to detect the hashing algorithm in use. Most clients MAY ignore the value, but if it is used, the client MUST verify the value matches the returned manifest. If the <reference> part of a manifest request is a digest, clients SHOULD verify the returned manifest matches this digest.

This project does already make use of `Docker-Content-Digest` [here](https://github.com/oras-project/oras-py/blob/a209c6cc145869e3056a7061398efca1c14f39d7/oras/layout/layout.py#L560) but fails to verify the value matches the returned manifest. This means that the project does not meet the 1.1.0 specification as this property is defined as a MUST.

The same applies for the blobs themselves as the spec makes clear:

> A GET request to an existing blob URL MUST provide the expected blob, with a response code that MUST be 200 OK. A successful response SHOULD contain the digest of the uploaded blob in the header Docker-Content-Digest. If present, the value of this header MUST be a digest matching that of the response body. Most clients MAY ignore the value, but if it is used, the client MUST verify the value matches the returned response body. Clients SHOULD verify that the response body matches the requested digest.

This project should verify both manifests and blobs against the digest that is:

1. requested
2. provided by the Docker-Content-Digest header.

judahrand:
We really _should_ be doing content verification when we download data from the registry. I believe other ORAS implementations do and certainly tools like Docker, Podman, etc. do. The OCI Spec(https://specs.opencontainers.org/image-spec/descriptor) says:

> Retrieved content SHOULD be verified against this digest when consumed via untrusted sources.

And then give guidance of what verification should check [here](https://github.com/opencontainers/image-spec/blob/main/descriptor.md?utm_source=chatgpt.com#verification).

Resolves #247

### DIFF
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -505,10 +505,10 @@ class Registry:
         digest_obj = oras.oci.Digest(digest)
         hasher = digest_obj.algorithm.hasher()
         blob_size = 0
-        staged = oras.utils.get_tmpfile()
+        outdir = os.path.dirname(outfile)
+        staged = oras.utils.get_tmpfile(tmpdir=outdir)
         try:
             # Ensure output directory exists first
-            outdir = os.path.dirname(outfile)
             if outdir and not os.path.exists(outdir):
                 oras.utils.mkdir_p(outdir)
             with self.get_blob(container, digest, stream=True) as r:
--- a/oras/provider.py
+++ b/oras/provider.py
@@ -3,18 +3,14 @@ __copyright__ = "Copyright The ORAS Authors."
 __license__ = "Apache-2.0"
 
 import copy
-import enum
-import hashlib
-import hmac
 import os
-import re
 import sys
 import urllib
 from contextlib import contextmanager, nullcontext
 from dataclasses import asdict
 from http.cookiejar import DefaultCookiePolicy
 from tempfile import TemporaryDirectory
-from typing import Any, Callable, Generator, List, Optional, Tuple, Union
+from typing import Callable, Generator, List, Optional, Tuple, Union
 
 import jsonschema
 import requests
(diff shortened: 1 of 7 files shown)


oras/provider.py
```diff
@@ -27,6 +31,87 @@
 from oras.types import container_type
 from oras.utils.fileio import PathAndOptionalContent
 
+_DIGEST_PATTERN = re.compile(
```
vsoch: These are not well placed - a Digest definition and helper functions do not belong in provider.py.

I'm guessing you asked the LLM to do this here. Let's simplify and just remove this final commit because it has a lot of problems.

vsoch: Oh this is a completely different PR.

These additions do not follow the conventions of the project. We have a specific module for types, and then functions to interact with them should be on the type. They do not belong in provider.py. Please check with the oras community about policy for using LLMs I'm not familiar - in other projects we have to add a line to the commit that code was assisted by Claude, etc.

judahrand: > I'm guessing you asked the LLM to do this here. Let's simplify and just remove this final commit because it has a lot of problems.

I actually asked an LLM to do it, didn't like what it did and rewrote it myself for the most part so this wasn't really LLM authored 😅

I'll happily move the types around 🤷 I do think this is an important addition, however. I was honestly a bit surprised the project wasn't already doing it as other ORAS implementations do.

vsoch: I'm not up to speed, so we will need to ask @FeynmanZhou.

judahrand: Were you thinking that the types should live in `oci.py`? I've reworked the PR to do this and also moved that blob validation inside `download_blob`. Maybe this is more how you think it should work?

Issue #229, opened 2026-02-09
Push outputs information text to stdout


vraamark:
The push function has a print that writes to stdout, which makes it difficult to use piping. My CLI tool, which uses Python Oras, sends all informational output to stderr and, at the end, returns a Digest to stdout. I would like to pipe the digest to another command, but due to the push issue, I cannot do that.

The line in question can be found here: https://github.com/oras-project/oras-py/blob/3c639859298a7c56e037f6c3c022584072f93f6a/oras/provider.py#L866

The example code suggests I can use a "quiet" parameter with the push command, but the parameter is not valid in the push source code.

https://github.com/oras-project/oras-py/blob/3c639859298a7c56e037f6c3c022584072f93f6a/examples/simple/push.py#L61

vsoch:
It should work in this case! I'd be happy to review a pull request to make changes so that it does.

PR #249, opened 2026-08-07
fix: route push completion output through logger

tomatotomata:
## What changed

`Registry.push()` now routes its completion message through the existing logger instead of writing directly to stdout. It also accepts the documented `quiet` option, so callers such as `examples/simple/push.py` can suppress that message without a `TypeError`.

## Why

The direct `print()` made it impossible to pipe a digest or other command output cleanly. This keeps informational output on the logger path and preserves the completion message by default.

## Validation

- `python -m pytest -p no:recording oras/tests/test_provider.py -k 'quiet_output or quiet_suppresses' -q` passed: 2 tests
- `python -m compileall -q oras` passed
- `python -m isort --check-only oras/provider.py oras/tests/test_provider.py` passed
- `git diff --check` passed
- Full suite: 54 passed, 19 skipped, 1 unrelated baseline failure in `test_sanitize_path` caused by Windows path separator expectations

This addresses #229. I used an AI coding assistant during implementation and reviewed the resulting diff and test evidence locally.


vsoch:
This looks good @tomatotomata - a simple and reasonable change. Please see the DCO requirements here: https://github.com/oras-project/oras-py/pull/249/checks?check_run_id=92784221176

tomatotomata:
The DCO check was the only action-required gate. I amended the head commit with the authorized sign-off `Signed-off-by: ahmad <ahmadalgaidy@hotmail.com>` and force-pushed the same branch. The code and validation scope are unchanged, so the checks can rerun against the new head.

vsoch (approved):

PR #250, opened 2026-08-18
Refactor/transport boundary

### LINKED ISSUE #242: Async support for the client
Hi, The client is sync-only right now — push()/pull() are blocking and everything goes through a requests.Session in do_request. I couldn't find this raised before, so wanted to ask before spending time on it.

My use case: I'm pushing/pulling a lot of artifacts concurrently from an async service, and today the only option is a thread pool with one Registry per thread. Native async/await would let this run on the existing event loop instead.

 Poking around, it looks like a bigger change than one function — the auth backends (token, ecr, …) do their own HTTP calls, @retry uses time.sleep, and ECR is on boto3, so it's not just do_request. 

Two ways I could see doing it:

 - A: an additive AsyncRegistry on top of httpx (sync API untouched), possibly using unasync so there's one source of truth.
 - B: keep it sync but add opt-in threaded concurrency internally.

 I'd lean towards A, but happy to go either way. I'd also want to know how this should fit around #240, since that reworks push/pull quite a bit.

 Is async something you'd accept in principle? If so, any preference on approach? Happy to put up a PR , just wanted to check direction first. 

Thanks!

chinmay1819:
## Summary

This PR addresses [#242](https://github.com/oras-project/oras-py/issues/242) by introducing the architectural foundation required for native asynchronous support in ORAS-py.

The implementation separates Registry/OCI behavior from HTTP transport concerns, allowing synchronous and asynchronous execution paths to share the same high-level logic while using transport implementations appropriate to each execution model.

The async API is designed as an explicit API rather than making existing synchronous methods dynamically behave as either synchronous or asynchronous.

## Issue

Closes #242

## Reviewer

@vsoch — Vanessa Sochat, please review the architectural direction and implementation.

## Architectural Strategy

The main design goal is to avoid maintaining two independent implementations of the Registry logic.

Instead of duplicating the existing `Registry` implementation into an `AsyncRegistry`, the architecture separates:

1. Registry / OCI semantics
2. HTTP transport
3. Sync execution
4. Async execution

The intended architecture is:

```text
                         Public API
                            │
                 ┌──────────┴──────────┐
                 │                     │
             Sync API              Async API
                 │                     │
                 └──────────┬──────────┘
                            │
                   Shared Registry /
                     OCI semantics
                            │
                  ┌─────────┴─────────┐
                  │                   │
           Sync Transport       Async Transport
                  │                   │
               requests              httpx

### DIFF
--- a/docs/getting_started/user-guide.md
+++ b/docs/getting_started/user-guide.md
@@ -688,6 +688,57 @@ client = oras.client.OrasClient(auth_backend="ecr")
 client.pull(target="123456789012.dkr.ecr.us-east-1.amazonaws.com/my-repo:latest")
 ```
 
+## Asynchronous Client
+
+There is an asynchronous client alongside the synchronous one. It needs `httpx`,
+which is an extra, so it is not installed unless you ask for it:
+
+```bash
+$ pip install oras[async]
+```
+
+The two clients are separate classes rather than one class with two modes, so
+whether a call is awaited is clear from the client you created. `AsyncOrasClient`
+owns a connection pool for its lifetime, so use it as a context manager, or call
+`aclose()` when you are done with it:
+
+```python
+import oras.client
+
+async with oras.client.AsyncOrasClient(hostname="localhost:5000", insecure=True) as client:
+    await client.push(files=["artifact.txt"], target="localhost:5000/dinosaur/artifact:v1")
+    files = await client.pull(target="localhost:5000/dinosaur/artifact:v1", outdir="pulled")
+```
+
+```python
+client = oras.client.AsyncOrasClient(hostname="localhost:5000", insecure=True)
+try:
+    manifest = await client.get_manifest("localhost:5000/dinosaur/artifact:v1")
+finally:
+    await client.aclose()
+```
+
+The registry operations are the asynchronous ones: `push`, `pull`, `get_manifest`,
+`get_manifest_content`, `upload_manifest`, `upload_manifest_content`, `upload_blob`,
+`download_blob`, `blob_exists`, `get_tags`, `delete_tag`, `delete_tags` and
+`do_request`. Reading and writing files on disk stays synchronous, because that is
+local work rather than something to wait on.
+
+Because a client holds one connection pool, operations on it can run together:
+
+```python
+import asyncio
+
+async with oras.client.AsyncOrasClient(hostname="localhost:5000", insecure=True) as client:
+    first, second = await asyncio.gather(
+        client.pull(target="localhost:5000/dinosaur/artifact:v1", outdir="one"),
+        client.pull(target="localhost:5000/dinosaur/other:v1", outdir="two"),
+    )
+```
+
+The synchronous client is unchanged and needs nothing new installed, so both APIs
+can be used from the same project.
+
 ## Custom Clients
 
 The benefit of Oras Python is that you can create a subclass that easily implements
--- a/oras/auth/__init__.py
+++ b/oras/auth/__init__.py
@@ -1,4 +1,4 @@
-import requests
+from oras.transport import Transport
 
 from .basic import BasicAuth
 from .ecr import EcrAuth
@@ -16,13 +16,33 @@ class AuthenticationException(Exception):
 
 
 def get_auth_backend(
-    name="token", session=None, insecure=False, tls_verify=True, **kwargs
+    name="token",
+    session=None,
+    insecure=False,
+    tls_verify=True,
+    transport=None,
+    **kwargs,
 ):
+    """
+    Get an auth backend by name, ready to send requests.
+
+    :param name: name of the backend to create
+    :type name: str
+    :param session: a session for the backend to use, if no transport is given
+    :type session: requests.Session
+    :param insecure: use http instead of https
+    :type insecure: bool
+    :param tls_verify: enable/disable tls verification or use a custom CA-Bundle
+    :type tls_verify: bool or str
+    :param transport: transport to send requests with, usually the provider's
+                      so that both share connections. One is created from the
+                      session and tls_verify when not provided
+    :type transport: oras.transport.Transport
+    """
     backend = auth_backends.get(name)
     if not backend:
         raise ValueError(f"Authentication backend {backend} is not known.")
     backend = backend(**kwargs)
-    backend.session = session or requests.Session()
+    backend.transport = transport or Transport(session=session, tls_verify=tls_verify)
     backend.prefix = "http" if insecure else "https"
-    backend._tls_verify = tls_verify
     return backend
--- a/oras/auth/ecr.py
+++ b/oras/auth/ecr.py
@@ -2,6 +2,7 @@ __author__ = "Vanessa Sochat"
 __copyright__ = "Copyright The ORAS Authors."
 __license__ = "Apache-2.0"
 
+import asyncio
 import re
 from typing import Optional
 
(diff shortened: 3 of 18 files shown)

### CODE REVIEW

**Comment by chinmay1819** [comment]
@vsoch Please let me know if any changes needed.

**Comment by vsoch** [comment]
I am out until September and then on travel, so your best bit is to ping someone in the oras slack to review. Thanks!

**Comment by chinmay1819** [comment]
@jdolitsky could you please review this pr and let me know if any changes are required ?

**Comment by chinmay1819** [comment]
Hi @vsoch , sorry to disturb you on your holidays, but I have tried with multiple people to review the pr on slack channel but as all of them are pre-occupied with work, no one is ready to review & approve the pr. Can you please take a look into this :)

**Review by TerryHowe — COMMENTED** [review]
Critical Issues

- [oras/provider_async.py:350] The async blob upload body is a one-shot generator and is silently re-sent empty on the auth retry. put_upload passes data=iter_file(blob, ...) into do_request, which re-sends the same data object on a 401 challenge, on the 403 refresh, and on every retry_async attempt. An async generator can only be consumed once, so the second send transmits zero bytes while still declaring Content-Length: layer["size"] and the digest= query param. I reproduced it with a fake transport that 401s the first PUT:

PUT .../uploads/1?digest=sha256:b94d27... body= b'hello world'
PUT .../uploads/1?digest=sha256:b94d27... body= b''      <-- auth retry

- A real registry will either hang on the unfulfilled Content-Length or reject on digest mismatch. This is a regression vs. sync, not a mirrored bug: oras/provider.py put_upload uses data=fd.read() (bytes), which is re-sendable. Fix by materializing the body, or by re-creating the iterator per attempt (e.g. pass a factory and call it inside the transport).
- [oras/provider_async.py:294] download_blob bypasses do_request, so it has no authentication-challenge handling and no retry. It calls self.transport.stream(...) directly and only attaches a header when isinstance(self.auth, TokenAuth) and self.auth.token is not None. Consequences: with BasicAuth no Authorization header is ever sent; with EcrAuth/TokenAuth before a token is cached, the response is a 401 that raise_for_status() turns into an exception instead of a challenge to answer; and a token scoped narrower than the blob endpoint can never be refreshed. Sync goes download_blob → get_blob → do_request, which handles all three. This is invisible in the common pull() path only because get_manifest happens to warm the token first.

Important Issues

- [oras/decorator.py:118] retry_async loses the sync fast-fail on TLS errors. Sync retry has except (requests.exceptions.SSLError, ImportError): raise; the async version keeps only ImportError. httpx surfaces certificate failures as httpx.ConnectError (a plain Exception), so a bad cert or wrong hostname now retries 5 times with 2+3**n backoff — roughly 48 s of sleeping — before the user sees the error. Add the httpx connect/SSL classes to the re-raise tuple (lazily, to keep httpx optional).
- [oras/transport_async.py:130] stream() never calls _forget_cookies(). The module docstring and _forget_cookies exist specifically for the Harbor CSRF case, but the streaming path — blob downloads, i.e. most of the traffic volume — leaves whatever the registry set sitting in the httpx jar for subsequent requests. Related: because _forget_cookies runs after request() returns, cookies are still honored across redirects within a single request, whereas the sync DefaultCookiePolicy(allowed_domains=[]) refuses them outright. Worth closing both gaps.
- DCO is failing and no test workflow has run. None of the five commits carry a Signed-off-by trailer, and the head SHA's only check is DCO: action_required — the auth-tests workflow has never executed on this branch. That means every async integration test (test_async_push_pull_*, test_async_chunked_upload, test_async_index_is_pushed_and_read_back) is currently unverified against a real registry, which is exactly where the two critical issues above would surface. Rebase with git rebase --signoff (and squash the Merge branch 'oras-project:main' commit) so CI can run.

**Comment by TerryHowe** [comment]
DCO failure as well, might want to squash the commits and signoff the one commit.

**Comment by chinmay1819** [comment]
Thanks for reviewing @TerryHowe . I will look into the mentioned issues :)

**Review by TerryHowe — APPROVED** [review]
Wrong PR, please ignore.

DCO issue here

**Review by TerryHowe — CHANGES_REQUESTED** [review]
DCO

**Comment by chinmay1819** [comment]
Hi @TerryHowe , I have fixed the issues that you mentioned in review of this pr. Now, DCO is also successful. Could you please review the changes and merge the pr ? Thank you :)

**Comment by chinmay1819** [comment]
@TerryHowe , (Gentle reminder) could you please review the pr and merge it ?

File: docs/getting_started/user-guide.md [hunk 1]
```diff
@@ -688,6 +688,57 @@ client = oras.client.OrasClient(auth_backend="ecr")
...
+try:
+    manifest = await client.get_manifest("localhost:5000/dinosaur/artifact:v1")
+finally:
+    await client.aclose()
+```
+
+The registry operations are the asynchronous ones: `push`, `pull`, `get_manifest`,
+`get_manifest_content`, `upload_manifest`, `upload_manifest_content`, `upload_blob`,
+`download_blob`, `blob_exists`, `get_tags`, `delete_tag`, `delete_tags` and
+`do_request`. Reading and writing files on disk stays synchronous, because that is
```
vsoch, line 724: `do_request` is not explicitly labeled as private (meaning, with an underscore) but it is considered private to the class. I would not list it here.

docs/getting_started/user-guide.md
```diff
@@ -688,6 +688,57 @@ client = oras.client.OrasClient(auth_backend="ecr")
...
+local work rather than something to wait on.
+
+Because a client holds one connection pool, operations on it can run together:
+
+```python
+import asyncio
+
+async with oras.client.AsyncOrasClient(hostname="localhost:5000", insecure=True) as client:
```
vsoch, line 732: Nice!

oras/auth/__init__.py
```diff
@@ -16,13 +16,33 @@ class AuthenticationException(Exception):
...
+    transport=None,
+    **kwargs,
 ):
+    """
+    Get an auth backend by name, ready to send requests.
+
+    :param name: name of the backend to create
+    :type name: str
+    :param session: a session for the backend to use, if no transport is given
+    :type session: requests.Session
```
vsoch, line 32: I like the abstraction of transport. The creation of transport.py and transport_asyc.py is a code smell that likely was created by an agent. It indicates (to me) there should be a transport module with a default and async class, meaning a directory with an init, async, and session, and with maximally shared structure in a base.py or similar.

oras/auth/ecr.py
```diff
@@ -60,28 +61,100 @@ def authenticate_request(
...
+
+        This is a blocking SDK call. Callers on an event loop run it in a
+        worker thread rather than awaiting it.
+
+        :param realm: the realm the token is for
+        :type realm: str
+        :param region: the AWS region to ask
+        :type region: str
+        """
+        try:
```
vsoch, line 134: I think we should keep this in init so we fail fast and do not require the client to be created freshly each attempt.

oras/auth/token.py
```diff
@@ -65,40 +72,107 @@ def authenticate_request(
...
 
-        logger.error(
-            "This endpoint requires a token. Please use "
-            "basic auth with a username or password."
-        )
-        return headers, False
+        return self._no_token(headers)
 
-    def request_token(self, h: auth_utils.authHeader) -> bool:
+    async def authenticate_request_async(self, original, headers: dict, refresh=False):
```
vsoch, line 140: Why do we need an async auth request, period? If we just have one auth request (non async) can we not use it in both? 

It seems a bit redundant, and that we have to maintain two versions of the same (or similar) operations. Is the async implementation much better or faster to warrant adding it? And if so (and we just choose one) can we not do asyncio.run in the non async clients?

oras/auth/token.py
```diff
@@ -163,19 +247,58 @@ def request_anonymous_token(self, h: auth_utils.authHeader) -> bool:
...
+        """
+        request = self._anonymous_token_request(h)
+        if not request:
+            return None
+
+        realm, params = request
+        response = self.transport.request(realm, "GET", params=params)
+        return self._token_from_response(response)
+
+    async def request_anonymous_token_async(
```
vsoch, line 289: Again, a lot of redundancy here. The async client makes sense, but I'm not sure that needs to extend to auth. If possible I'd prefer just one strategy shared by both.

oras/transport.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+
+__copyright__ = "Copyright The ORAS Authors."
+__license__ = "Apache-2.0"
+
+from http.cookiejar import DefaultCookiePolicy
+from typing import IO, Optional, Union
+
+import requests
+
```
vsoch, line 26: Already stated, but this would be good as a transport (directory) module with scoped files.

oras/version.py
```diff
@@ -2,7 +2,7 @@
 __copyright__ = "Copyright The ORAS Authors."
 __license__ = "Apache-2.0"
 
-__version__ = "0.2.43"
+__version__ = "0.2.51"
```
vsoch, line 5: Why the huge jump? Did I miss something?

vsoch (commented):
A first review to start. I have more detailed comments but will save for next round. Thanks!

PR #254, opened 2026-09-17
Fix temporary archive cleanup when pulling directories

### LINKED ISSUE #246: Pulling an artifact with media type `application/vnd.oci.image.layer.v1.tar+gzip` leaks a temporary file

Shubham-Padkonde:
Pulling a gzip-compressed directory leaves its temporary archive and containing directory on disk, including after a failed download or extraction. Repeated pulls can therefore accumulate disk usage.

Scope the archive download and extraction to a `TemporaryDirectory`, using the same temporary-file pattern as `temporary_empty_config`. The extracted output remains available, including when `pull` chooses its output directory automatically. Cleanup also runs when downloading or extracting raises, and the original exception still propagates.

Fixes #246. Thanks to @judahrand for the diagnosis and earlier work in #244; this is a smaller change limited to the directory-pull path, without changing the shared temporary-file helper API.

Validation:
- Six regression cases fail on the original implementation and pass with the fix: successful real archive extraction, interrupted download, and invalid archive, each with explicit and automatic output directories. They check archive/directory cleanup and retained extracted contents.
- Linux, Python 3.11: `python -m pytest oras/ -q` — 61 passed, 19 skipped. Registry-dependent tests were skipped because no local registry is running.
- Windows, Python 3.12: the six new cases pass; the existing extraction helper emits two tar-filter deprecation warnings.
- `pre-commit run --all-files` passes all configured checks.

Prepared with OpenAI Codex assistance. No live registry integration test was run.


oras/tests/test_provider.py
```diff
@@ -18,6 +21,63 @@
...
+                    "digest": "sha256:example",
+                    "annotations": {oras.defaults.annotation_title: "artifact"},
+                }
+            ]
+        },
+    )
+    downloads = []
+    payload = b"artifact contents"
+
+    def download_blob(container, digest, destination):
```
vsoch, line 49: Can you better use existing library logic? My concern is that this does not mirror what oras would do, and we have to update two places for any changes.

vsoch (commented):
Looks good! This is a simple, logical fix. One comment and then we are good to go.

oras/tests/test_provider.py
Shubham-Padkonde: Good point. In 57fd5d1 the test builds the artifact the way push does: a real directory archived with `oras.utils.make_targz`, described with `oras.oci.NewLayer(..., is_dir=True)` plus the title annotation, and the stubbed `download_blob` now copies that archive (and checks the layer digest). The six cases still fail without the fix and pass with it.

oras/tests/test_provider.py
```diff
@@ -27,45 +27,39 @@ def test_pull_cleans_temporary_archive(
...
-            archive.addfile(member, io.BytesIO(payload))
+        shutil.copyfile(archive, destination)
 
     monkeypatch.setattr(client, "download_blob", download_blob)
     outdir = None if use_default_outdir else str(tmp_path / "output")
     if outcome == "success":
         files = client.pull("registry.example/repository:tag", outdir=outdir)
         assert len(files) == 1
-        assert (Path(files[0]) / "content.txt").read_bytes() == payload
+        assert (Path(files[0]) / "content.txt").read_text() == "artifact contents"
```
vsoch: Can we make this a variable instead of a repeated string?

Shubham-Padkonde: Done in e9a0f66: the test now binds `target`, `content_name` and `content` once and uses them for the artifact it creates, both `client.pull(...)` calls and the assertion.

Shubham-Padkonde: Correction: the commit is bce6882 (I re-made it with the sign-off DCO wanted); the change is the same.

